PR #111, opened 2025-11-17
dark mode implementation

ptrjk:
Hi, I implemented dark mode to your application, I hope you will like it, it took some time for me to do it, and I learned new things!

thank you


DenizAltunkapan:
@ptrjk your darkmode implementation looks very good!
image
image

Please format your code properly with the `npx prettier --write .` command in the frontend directory, so that the workflows don't fail anymore.
After that, we should be able to merge this PR : )

Issue #146, opened 2025-12-23
[Feature]: Introduce common_problems.md for recurring setup issues


DenizAltunkapan:
## Problem

As the project is used across different operating systems and environments, recurring setup and startup issues have appeared that are not immediately obvious to new contributors. These issues often lead to duplicated questions or GitHub issues and increase onboarding friction.

## Proposed solution

Introduce a dedicated documentation file `common_problems.md` that collects common, well-understood setup and startup problems along with brief explanations and clear, actionable solutions.  
This document should complement the README by keeping troubleshooting details out of the main setup instructions.

## Acceptance criteria

- [ ] A new file `common_problems.md` is added to the repository
- [ ] The document contains at least one documented recurring issue with cause and solution
- [ ] The README references `common_problems.md` as a troubleshooting resource

PR #147, opened 2025-12-23
Make JDBC timezone configurable and document JVM startup workaround

### LINKED ISSUE #145: Issue: Backend startup fails due to deprecated JVM timezone during JDBC handshake
**Description**
On some systems, the backend fails to start because PostgreSQL rejects a deprecated JVM timezone identifier during the initial JDBC connection handshake.

The PostgreSQL JDBC driver propagates the JVM’s default timezone as a startup parameter. On certain JVM / OS combinations, this defaults to `Asia/Calcutta`, which recent PostgreSQL versions reject in favor of `Asia/Kolkata`, resulting in:

`2025-12-23T13:30:21.698+05:30 ERROR 14408 --- [backend] [  restartedMain] o.h.engine.jdbc.spi.SqlExceptionHelper   : FATAL: invalid value for parameter "TimeZone": "Asia/Calcutta"`

Because the failure occurs during connection negotiation, Hibernate cannot obtain JDBC metadata, causing a cascade of secondary startup errors.

**Environment**
Java 21+
PostgreSQL (notably recent Docker images)
Startup via `./mvnw spring-boot:run`
OS-dependent (Windows/Linux/macOS)

**Notes**
This occurs before Hibernate executes any queries
`hibernate.jdbc.time_zone` does not affect JDBC startup parameters
Root cause is the JVM → JDBC → PostgreSQL handshake

**Proposed Direction**
Avoid reliance on JVM default timezone values at startup
Default to UTC where appropriate
Allow explicit timezone configuration via JVM arguments or environment configuration
Document cross-platform startup guidance for contributors

**Expected Outcome**
Backend starts reliably across environments
No PostgreSQL startup failures due to deprecated timezone identifiers

aditya-7562:
## Summary

Fix backend startup failures caused by deprecated JVM timezone identifiers being propagated to PostgreSQL during the initial JDBC handshake.

This PR makes JDBC timestamp handling deterministic and documents how to explicitly configure the JVM timezone at startup to avoid relying on OS/JDK defaults.

## Linked issue

Closes #145

## How to test

1. Start PostgreSQL using Docker:

   ```bash
   docker compose up -d
   ```

2. Run the backend normally:

   ```bash
   ./mvnw spring-boot:run
   ```

3. If your environment defaults to a deprecated timezone such as `Asia/Calcutta`, run the backend with an explicit JVM timezone:

   ```bash
   ./mvnw "-Dspring-boot.run.jvmArguments=-Duser.timezone=UTC" spring-boot:run
   ```

4. Verify:

   * Backend starts successfully
   * No `FATAL: invalid value for parameter "TimeZone"` error
   * Swagger UI loads at `http://localhost:8080/swagger-ui.html`

## Notes / Risk

* No breaking changes
* Defaults to UTC for JDBC timestamp handling
* Does not hard-code regional timezone values
* Hibernate configuration affects timestamp semantics after connection establishment
* README guidance covers JVM-level startup behavior where the failure occurs

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,28 +92,14 @@ cd backend
 
 ### Timezone configuration (important for some environments)
 
-On some systems, the JVM may default to deprecated timezone identifiers (for example `Asia/Calcutta`), which PostgreSQL rejects during JDBC connection initialization.
+On some systems, the backend may fail to start because PostgreSQL rejects a deprecated JVM timezone identifier during startup.
 
-If the backend fails to start with an error similar to:
+If you encounter an error like:
 
 ```text
 FATAL: invalid value for parameter "TimeZone"
 ```
-
-explicitly configure the JVM timezone when running the backend. This ensures the PostgreSQL JDBC driver does not propagate deprecated timezone identifiers during connection startup.
-
-Recommended (cross-platform, works on Windows, macOS, Linux)
-
-```bash
-./mvnw "-Dspring-boot.run.jvmArguments=-Duser.timezone=UTC" spring-boot:run
-```
-
-Alternative (Unix-like shells)
-
-```bash
-JAVA_OPTS=-Duser.timezone=UTC
-./mvnw spring-boot:run
-```
+see common_problems.md for platform-specific troubleshooting and startup guidance.
 
 ---
 


DenizAltunkapan (changes requested):
One small suggestion regarding the README: the new timezone section is very thorough, but it might be a bit long for a primary setup guide. Since this is more of a platform-specific startup pitfall, it could fit nicely into a dedicated common_problems.md (as discussed in #146), with a short reference from the README instead.

Alternatively, we could keep a much shorter note here and link to the more detailed explanation. That would help keep the README concise while still documenting the issue properly.

Let me know what you think

DenizAltunkapan (approved):
thanks @aditya-7562

PR #151, opened 2025-12-26
Configure automatic database creation via init script

MetehanTosun:
- Add init-db.sql to create cloud-db on container startup
- Update docker-compose.yml to mount init script

## Summary

Added automatic database creation for both `vaultdb` and `cloud-db` via PostgreSQL init script. Previously, `cloud-db` had to be created manually in pgAdmin, which caused startup failures for the cloud-page application.

## How to test
1. Stop and remove existing containers: `docker compose down -v`
2. Start containers: `docker compose up -d`
3. Verify both databases exist:
```bash
   docker exec -it vault-web-db-1 psql -U metehan -l
```
   You should see both `vaultdb` and `cloud-db` in the list.
4. Start both backend applications - they should connect successfully without manual database creation.

## Notes / Risk

- Requires volume deletion (`-v` flag) for init script to run on existing setups
- Init script only runs on first container startup with empty volumes
- Low risk: Only affects local development environment, not production

### DIFF
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@
 DB_PORT=5433
 POSTGRES_USER=your_username
 POSTGRES_PASSWORD=your_password
-POSTGRES_DB=vaultdb
+POSTGRES_DB=vault-db
 
 # pgAdmin config
 PGADMIN_DEFAULT_EMAIL=example@email.com


DenizAltunkapan (approved):
@MetehanTosun Thank you for your contribution

Issue #189, opened 2026-02-04
[Feature]: User Search / Filtering on Home Page


DenizAltunkapan:
## Problem
Currently, the home page displays all users in the network, which does not scale well as the number of users grows. It should be possible to easily search or filter users to quickly find the intended chat partner in the home page.

Add a search input to the home page that allows filtering users by username. The initial implementation can be client-side, but the search logic should be flexible enough (parametized inside the application properties for example) to support fuzzy matching (e.g. Levenshtein distance with a configurable threshold) or more performant algorithms in the future to handle typos and partial matches.

This will significantly improve usability and align the UI with common chat application patterns.

Navila48:
@DenizAltunkapan Can you please review the PR?

Navila48:
Since PR is merged and feature is already implemented, I think this issue can be closed @DenizAltunkapan

Navila48:
@DenizAltunkapan One small request. Here, my PR is not displayed with the issue. Could you please link the corresponding PR with this issue?

<img width="931" height="190" alt="Image" src="https://github.com/user-attachments/assets/2b527d3c-291b-4a35-b38c-4f5a933e3bb9" />

PR #215, opened 2026-04-12
Feature/ Add support for searching users on home page

Navila48:
## Summary

- Instant client-side search
- Scalable foundation for backend search later

## Linked issue
[Issue 189](https://github.com/Vault-Web/vault-web/issues/189)

Screenshot:
We have these 3 users:
image

After searching for user2:

image

### DIFF
--- a/frontend/src/app/pages/home/home.component.html
+++ b/frontend/src/app/pages/home/home.component.html
@@ -52,7 +52,7 @@
   </div>
 
   <div *ngIf="!isLoading && !error">
-    <div *ngIf="privateChats.length > 0">
+    <div *ngIf="filteredPrivateChats.length > 0">
       <div
         class="section-heading px-4 py-2 text-xs font-semibold text-gray-600 uppercase"
       >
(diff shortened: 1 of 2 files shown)


Navila48:
@DenizAltunkapan This PR is waiting for your review :(

DenizAltunkapan:
@Navila48 
Could you also apply the search filtering to private chats, not just the “All Users” section?

In another issue, “All Users” will be collapsed by default. In that case, searching would only update a collapsed section, meaning the user wouldn’t see any visible changes, which can be confusing.

Filtering private chats as well would ensure that search results are always visible and consistent across the UI.

Navila48:
> Could you also apply the search filtering to private chats, not just the “All Users” section?

Thanks for sharing your insights. I was in doubt to add this search option also for private chats or not!
Done
image

**Another edge case:** 
When user/client search for a non-existing user both on chats and Users list, it will display a meesage instead of a blank page.
image

DenizAltunkapan (approved):

DenizAltunkapan:
@Navila48 Thank you for your contribution :)

PR #231, opened 2026-06-10
feat: show folder sizes in cloud view (consume /folders/content)

### LINKED ISSUE #230: [Feature]: Show folder sizes in cloud view (consume /folders/content)
## Context

[Vault-Web/cloud-page#75](https://github.com/Vault-Web/cloud-page/pull/75) (merged) makes the backend compute folder sizes: folders now report the **recursive sum of the regular files they contain** via `GET /folders/content` (`FolderContentItemDto.size`), instead of always `0`.

However, this is **not yet visible in the vault-web frontend**. Two reasons:

1. **The cloud view doesn't use the changed endpoint.** `CloudService` (`frontend/src/app/services/cloud.service.ts`) calls `GET /folders` and `GET /folders/path` (returning `FolderDto` with `FolderListItemDto[]`), not the paginated `GET /folders/content` that PR #75 touched.
2. **The folder DTO it uses has no `size` field.** `FolderListItemDto` only has `name`, `path`, `directChildrenCount`, `lastModifiedAt`. In `cloud.component.ts`, the **Size** column for folders is rendered as the item count (`` `${directChildrenCount} items` ``), not bytes. Only files use `formatFileSize(...)`.

So merging the backend PR has no visible effect until the frontend is wired up.

## Goal

Surface folder storage sizes in the cloud view, using the size already provided by `/folders/content`.

## Proposed approach (recommended)

Migrate the cloud view to the paginated `GET /folders/content` endpoint (`FolderContentItemDto`), which already provides `size` — plus pagination and sorting. Then render the folder size via `formatFileSize(...)` in the Size column.

Why this over just adding `size` to the `/folders` response:
- Avoids duplicating the recursive size-calculation logic into a second backend endpoint.
- `/folders/content` is the more capable endpoint (pagination + sorting + sizes); `/folders` loads everything at once and scales poorly.
- Pagination bounds the size-walk cost (only current-page folders are walked) — relevant to the performance follow-up tracked in cloud-page#62.

## Tasks

- [ ] Switch the cloud view / `CloudService` to `GET /folders/content` (pagination + sorting)
- [ ] Add a `FolderContentItemDto` model including `size`
- [ ] Render folder size via `formatFileSize(...)` in the Size column (instead of / alongside the item count)
- [ ] Verify: empty folders show `0 Bytes`, nested sizes sum correctly

## References

- Backend PR (merged): Vault-Web/cloud-page#75
- Original feature issue: Vault-Web/cloud-page#55
- Performance follow-up: Vault-Web/cloud-page#62

GabrielBBaldez:
Closes #230.

Thanks for opening this follow-up, @DenizAltunkapan — happy to carry the feature through to the UI.

## What

The cloud view rendered the **Size** column for folders as an item count (`N items`), because it consumed `GET /folders` whose `FolderListItemDto` has no `size`. This switches the entry listing to the paginated `GET /folders/content` endpoint — which reports each folder's recursive size since Vault-Web/cloud-page#75 — and renders it with `formatFileSize(...)`, same as files.

## Changes

- **`CloudService`** — new `getFolderContent(relativePath, page, size, sort)` calling `GET /folders/content`.
- **Models** — `FolderContentItemDto` (with `size`) and `PageResponseDto<T>`, mirroring the backend DTOs.
- **`CloudComponent`** — entries are now built from the content page (folders and files come in one list with a `directory` flag); folder rows show their recursive size via `formatFileSize(...)`. Navigation, breadcrumbs and file operations keep using the existing folder endpoints, so behavior there is unchanged.
- **Template** — table wired to PrimeNG lazy pagination (`onLazyLoad`); the paginator only appears when the folder has more entries than one page. Pagination keeps the backend size-walk bounded to the current page, which ties into the cloud-page#62 performance concern.

## Verification

- `npm run build` — succeeds (strict template type-checking included).
- `prettier --check` on the changed files — clean.
- Empty folders render `0 Bytes` via the existing `formatFileSize(0)` branch; nested folder sizes come straight from the backend's recursive sum (covered by tests in cloud-page#75).

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -68,6 +68,7 @@
       [lazy]="true"
       [paginator]="totalElements > pageSize"
       [rows]="pageSize"
+      [first]="contentFirst"
       [totalRecords]="totalElements"
       (onLazyLoad)="onPageChange($event)"
       styleClass="cloud-table text-xxs sm:text-xs"
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -65,6 +65,11 @@
   <section *ngIf="currentFolder && !loading" class="cloud-table-wrap">
     <p-table
       [value]="entries"
+      [lazy]="true"
+      [paginator]="totalElements > pageSize"
+      [rows]="pageSize"
+      [totalRecords]="totalElements"
+      (onLazyLoad)="onPageChange($event)"
       styleClass="cloud-table text-xxs sm:text-xs"
       [tableStyle]="{ 'min-width': '100%' }"
     >
--- a/frontend/src/app/models/dtos/FolderContentItemDto.ts
+++ b/frontend/src/app/models/dtos/FolderContentItemDto.ts
@@ -1,0 +1,8 @@
+export interface FolderContentItemDto {
+  name: string;
+  path: string;
+  directory: boolean;
+  size: number;
+  mimeType: string | null;
+  lastModifiedAt: number;
+}
--- a/frontend/src/app/models/dtos/PageResponseDto.ts
+++ b/frontend/src/app/models/dtos/PageResponseDto.ts
@@ -1,0 +1,6 @@
+export interface PageResponseDto<T> {
+  content: T[];
+  totalElements: number;
+  totalPages: number;
+  pageNumber: number;
+}
--- a/frontend/src/app/pages/cloud/cloud.component.ts
+++ b/frontend/src/app/pages/cloud/cloud.component.ts
@@ -81,6 +81,7 @@ export class CloudComponent implements OnInit {
   pageSize = 50;
   totalElements = 0;
   private contentPage = 0;
+  private contentRequestId = 0;
 
   private draggedPath: string | null = null;
   private draggedIsFolder = false;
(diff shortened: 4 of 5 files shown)


GabrielBBaldez:
Went through the Copilot review:

- **Stale responses on fast navigation (applied):** good catch — `loadFolderContent` now tags each request with an increasing id and ignores responses that are no longer the latest, so quick folder/page switches can't overwrite newer state with older data.
- **Folder size as `0 Bytes`:** kept as-is intentionally — showing folder sizes (including `0 Bytes` for empty ones) is the goal of #230, and the issue's verification list explicitly calls for empty folders rendering `0 Bytes`.
- **`totalItemsInView` returning the folder total:** kept — the label reads "items **in this folder**", so under pagination the folder total is the accurate value (the previous value only matched because everything was loaded at once).
- **`event.first` array handling:** kept — PrimeNG's `TableLazyLoadEvent.first` is typed `number | number[] | undefined`, and the strict template check requires handling both shapes.

`npm run build` and `prettier --check` still clean.

frontend/src/app/pages/cloud/cloud.component.html
```diff
@@ -65,6 +65,11 @@ <h1 class="cloud-title text-base sm:text-lg">Cloud Page</h1>
   <section *ngIf="currentFolder && !loading" class="cloud-table-wrap">
     <p-table
       [value]="entries"
+      [lazy]="true"
+      [paginator]="totalElements > pageSize"
+      [rows]="pageSize"
+      [totalRecords]="totalElements"
+      (onLazyLoad)="onPageChange($event)"
```
DenizAltunkapan, line 73: No [first] binding, so PrimeNG keeps its pagination offset across folder navigation. Going from a high page into a smaller folder can leave a stale page index and request an out-of-range page → blank listing. Please bind [first]="contentFirst" and reset it to 0 in loadRootFolder/navigateToFolder.

DenizAltunkapan:
@GabrielBBaldez please check my comment, after that the pr is ready to merge : )

GabrielBBaldez:
Good catch @DenizAltunkapan — applied in 7d59a6f: the table now binds `[first]="contentFirst"`, it resets to 0 in `loadRootFolder`/`navigateToFolder`, and `onPageChange` keeps it in sync when paging — so entering a smaller folder from a high page can't request an out-of-range page anymore. `npm run build` + prettier still clean.

DenizAltunkapan (approved):
@GabrielBBaldez Thank you very much. Looks good!

PR #238, opened 2026-06-14
feat: add trash UI for cloud view (consume /files/trash endpoints)

### LINKED ISSUE #237: [Feature]: Trash UI for cloud view (consume /files/trash endpoints)
## Context

[Vault-Web/cloud-page#79](https://github.com/Vault-Web/cloud-page/pull/79) adds a per-user **trash (soft delete)** on the backend: `DELETE /files` now moves a file into a per-user trash instead of deleting it permanently, plus new endpoints under `/files/trash` to list, restore, and permanently delete entries. Expired entries are auto-purged after 30 days.

This is **not yet visible in the vault-web frontend** — the cloud view has no way to view or manage trashed files, so the feature has no user-facing effect until it is wired up.

## Goal

Add a trash view to the cloud page that consumes the new endpoints.

## New backend endpoints (from cloud-page#79)

- `GET /files/trash` — list entries (`id`, `name`, `originalPath`, `deletedAt`, `size`)
- `POST /files/trash/{id}/restore` — restore to original path
- `DELETE /files/trash/{id}` — permanently delete

## Tasks

- [ ] Add trash methods to `CloudService` (`frontend/src/app/services/cloud.service.ts`) for the three endpoints
- [ ] Add a `TrashEntry` model (`id`, `name`, `originalPath`, `deletedAt`, `size`)
- [ ] Trash view: list entries with name, original path, deleted date and size
- [ ] Restore and permanent-delete actions (confirm dialog on permanent delete)
- [ ] Entry point from the cloud view (e.g. a "Trash" / recycle-bin link)
- [ ] Handle the restore conflict case gracefully (a file already exists at the original path)

## References

- Backend PR: Vault-Web/cloud-page#79
- Original feature issue: Vault-Web/cloud-page#60

GabrielBBaldez:
Closes #237.

Adds the cloud-page trash view that consumes the per-user trash endpoints from Vault-Web/cloud-page#79.

## What's included
- `CloudService` methods for the three endpoints (`GET /files/trash`, `POST /files/trash/{id}/restore`, `DELETE /files/trash/{id}`) plus a `TrashEntryDto` model (`id`, `name`, `originalPath`, `deletedAt`, `size`).
- A standalone `TrashComponent` (route `/cloud/trash`, behind `authGuard`): a table of trashed entries (name, original location, deleted date, size) with **Restore** and **permanent-delete** actions; permanent delete asks for confirmation. Loading / error / empty states included.
- A **Trash** entry point in the cloud toolbar.
- Restore-conflict handling: when a file already exists at the original path, the user gets a clear message instead of a generic failure.

## Notes
- Depends on Vault-Web/cloud-page#79 (the backend trash endpoints) being merged for the endpoints to exist.
- The backend currently surfaces a restore conflict as a 500 (`FileAlreadyExistsException` isn't mapped to 409); the UI handles both that and a future 409, so it degrades gracefully either way. Happy to send a one-line follow-up on cloud-page#79 mapping it to 409 for a cleaner contract.
- Verified locally with `ng build` (AOT + strict templates) and Prettier.

### DIFF
--- a/frontend/src/app/app.routes.ts
+++ b/frontend/src/app/app.routes.ts
@@ -4,6 +4,7 @@ import { authGuard } from './auth.guard';
 import { HomeComponent } from './pages/home/home.component';
 import { RegisterComponent } from './pages/register/register.component';
 import { CloudComponent } from './pages/cloud/cloud.component';
+import { TrashComponent } from './pages/cloud/trash/trash.component';
 import { DashboardComponent } from './pages/dashboard/dashboard.component';
 import { PasswordManagerComponent } from './pages/password-manager/password-manager.component';
 
@@ -23,5 +24,6 @@ export const routes: Routes = [
   { path: 'password-manager', redirectTo: 'passwords', pathMatch: 'full' },
   { path: 'register', component: RegisterComponent },
   { path: 'cloud', component: CloudComponent, canActivate: [authGuard] },
+  { path: 'cloud/trash', component: TrashComponent, canActivate: [authGuard] },
   { path: '**', redirectTo: 'login' },
 ];
--- a/frontend/src/app/models/dtos/TrashEntryDto.ts
+++ b/frontend/src/app/models/dtos/TrashEntryDto.ts
@@ -1,0 +1,7 @@
+export interface TrashEntryDto {
+  id: string;
+  name: string;
+  originalPath: string;
+  deletedAt: string;
+  size: number;
+}
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -8,13 +8,22 @@
         {{ totalItemsInView }} items in this folder
       </p>
     </div>
-    <p-button
-      label="Refresh"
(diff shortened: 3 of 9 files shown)


GabrielBBaldez:
Addressed the Copilot comments in `444d351`:

- `formatFileSize` clamps the unit index (and adds TB/PB) so sizes ≥ 1 TB render correctly
- aria-label on the icon-only permanent-delete button
- moved the header-actions inline styles into the component stylesheet

Left two as-is, for consistency with the existing code:
- **Lazy-loading the Trash route** — every route in `app.routes.ts` is eagerly imported today, so making only this one lazy would be inconsistent. Happy to switch if you'd like to move the app toward lazy routes more broadly.
- **Explicit `: void` on `goToTrash`** — the sibling methods in `cloud.component.ts` don't declare return types, so I kept it consistent with the file.

Separately, restore conflicts now get a clean `409` from the backend (Vault-Web/cloud-page#79), which this UI already handles.

DenizAltunkapan (approved):
@GabrielBBaldez Thank you very much 😄

PR #239, opened 2026-06-14
fix: redirect to login on expired session

### LINKED ISSUE #232: [Bug]: Expired token shows a toast on navigation instead of redirecting to login
When a user's token expires, Vault Web doesn't reliably send them back to the login page. Navigation is still allowed, and depending on which request fails the user only sees an error toast while staying on a broken, unauthenticated page instead of being redirected to log in again.

There are two root causes. First, the route guard only checks whether a token *exists* in local storage, not whether it is still valid — `isLoggedIn()` in `auth.service.ts` is just `!!localStorage.getItem('token')`, so an expired-but-present token still counts as "logged in" and `authGuard` never redirects. Second, the token interceptor handles 401s inconsistently: when it attaches the token itself it attempts a refresh and logs the user out on failure (the correct path), but when a request already carries an `Authorization` header it simply passes the 401 through without refreshing or redirecting. On top of that, some components (e.g. the password manager and the E2EE service) catch 401s locally and show a toast without triggering the global logout.

Expected behavior: an expired session should consistently either refresh the token transparently or, if the refresh fails, clear the session and redirect to `/login` — regardless of which request triggered the 401. The guard should also treat an expired token as logged-out (e.g. by decoding the JWT `exp`) so a stale token can't grant access during navigation.

Relevant code:
- `frontend/src/app/services/auth.service.ts` — `isLoggedIn()` checks token presence only, not expiry
- `frontend/src/app/auth.guard.ts` — relies on `isLoggedIn()`, so it never redirects a stale-token user
- `frontend/src/app/core/interceptors/token.interceptor.ts` — the explicit-`Authorization` branch passes the 401 through without a refresh/logout
- `frontend/src/app/pages/password-manager/password-manager.component.ts` and `frontend/src/app/services/e2ee.service.ts` — handle 401 locally with a toast, bypassing the global logout flow

GabrielBBaldez:
Closes #232.

An expired session now consistently returns the user to login instead of leaving them on a broken, unauthenticated page:

- `isLoggedIn()` decodes the JWT `exp`, so `authGuard` treats a present-but-expired token as logged out and redirects.
- The token interceptor now handles `401`s the same way on both paths — a request that already carries an `Authorization` header also refreshes-and-retries once, or clears the session and redirects when the refresh fails (previously it passed the `401` through untouched). Since every `401` now flows through this path, components no longer get stuck showing a local error without the global logout firing.

Verified locally with `ng build` (AOT + strict templates) and Prettier.

### DIFF
--- a/frontend/src/app/core/interceptors/token.interceptor.ts
+++ b/frontend/src/app/core/interceptors/token.interceptor.ts
@@ -1,4 +1,9 @@
-import { HttpInterceptorFn, HttpErrorResponse } from '@angular/common/http';
+import {
+  HttpErrorResponse,
+  HttpHandlerFn,
+  HttpInterceptorFn,
+  HttpRequest,
+} from '@angular/common/http';
 import { inject } from '@angular/core';
 import { catchError, switchMap, throwError, ReplaySubject } from 'rxjs';
 import { AuthService } from '../../services/auth.service';
@@ -21,11 +26,7 @@ export const tokenInterceptor: HttpInterceptorFn = (req, next) => {
         }
 
         if (error.status === 401) {
-          // If the refresh token request itself fails, we must logout.
-          if (req.url.includes('/refresh')) {
-            authService.logout();
-          }
-          // Otherwise, just pass the 401 through.
+          return handleUnauthorized(req, next, authService, error);
         }
         return throwError(() => error);
       }),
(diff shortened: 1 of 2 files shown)


GabrielBBaldez:
Addressed the review in 398bfaa:

- JWT payload is now decoded as UTF-8 (base64url → bytes → `TextDecoder`), so non-ASCII claims parse reliably
- an undecodable/malformed token is treated as logged-out (fail closed) instead of reaching protected pages — with the robust decode this only triggers on genuinely malformed tokens
- the refresh-endpoint check matches `/auth/refresh` instead of any URL containing `/refresh`

Left the retry's `Authorization` handling as-is: every authenticated request here uses the session Bearer token, so replacing it on retry is correct — happy to gate it on the Bearer scheme if a non-Bearer `Authorization` is ever introduced.

DenizAltunkapan (approved):
lgtm, thanks!

PR #250, opened 2026-06-25
fix: keep chat composer visible when mobile keyboard opens

### LINKED ISSUE #214: [Bug]: Mobile layout bug in private chats: message input and send button hardly visible
In mobile view, the private chat screen sometimes does not show fully the message composer area (text input + send button).
As a result, users can read messages but cannot send new ones.

Please investigate and fix the responsive layout so that:

The input field and send button are always visible on mobile.
The composer remains accessible when the keyboard opens.
The fix works across common mobile viewport sizes and orientations.

29shivam:
## Summary
- The private chat overlay sized itself with `h-screen` (`100vh`), which tracks the static layout viewport and does not shrink when a mobile on-screen keyboard opens. Since the message list and input row are flex siblings inside a `position: fixed` shell, the input row got pushed below the visible area with no way to scroll it into view.
- Switched to `h-dvh` so the shell tracks the dynamic viewport.
- Added `interactive-widget=resizes-content` to the viewport `<meta>` tag so Chromium-based browsers shrink the layout viewport (and therefore `dvh`) when the keyboard appears, instead of leaving it covered.

Fixes #214

## Test plan
- [x] `ng build --configuration production` succeeds; confirmed compiled CSS contains `.h-dvh{height:100dvh}`
- [x] `eslint` clean on changed files
- [ ] Manual check on a real mobile device / Chrome DevTools device toolbar with keyboard open across a couple of viewport sizes and orientations

### DIFF
--- a/frontend/src/app/pages/private-chat-dialog/private-chat-dialog.component.html
+++ b/frontend/src/app/pages/private-chat-dialog/private-chat-dialog.component.html
@@ -1,4 +1,4 @@
-<div class="chat-shell fixed top-0 left-0 w-screen h-screen flex flex-col z-50">
+<div class="chat-shell fixed top-0 left-0 w-screen h-dvh flex flex-col z-50">
   <!-- Header -->
   <div
     class="chat-header px-4 py-3 border-b font-semibold flex justify-between items-center"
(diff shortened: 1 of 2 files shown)


DenizAltunkapan (approved):
@29shivam thank you

29shivam:
@DenizAltunkapan Thank you! looking forward towards more contribution.

PR #254, opened 2026-07-02
fix(cloud): prevent layout overflow on mobile viewports

### LINKED ISSUE #252: [Bug]: CloudPage layout breaks on file upload (mobile)
## Description

On mobile, the CloudPage layout breaks during file upload. Columns (`Size`, `Type`) get cut off and the file browser/breadcrumb area overflows horizontally instead of fitting the viewport.

Image

## Steps to Reproduce

1. Open the CloudPage on a mobile device (narrow viewport)
2. Navigate into a folder and upload a file
3. Observe the layout — table columns are clipped and content overflows the screen

## Expected Behavior

The layout stays responsive on mobile and content fits the viewport during and after upload.

29shivam:
## Summary

Fixes #252 — CloudPage table breaks on mobile due to missing responsive styles.

**Root cause:** `.cloud-table-wrap` had `overflow: hidden` which clips the 5-column table horizontally on narrow viewports instead of allowing a scroll or hiding non-essential columns.

**Changes:**
- `overflow: hidden` → `overflow-x: auto / overflow-y: hidden` on `.cloud-table-wrap` so the table can scroll horizontally as a fallback
- Hide **Size** and **Type** columns below 640 px — least essential metadata on small screens
- Hide **Modified** column below 420 px for very small devices
- Wrap `.cloud-headline` at 768 px so title + action buttons don't overflow
- Shrink `.cloud-entry-name` max-width at 768 px / 420 px to fit the narrower Name cell

## Test plan

- [ ] Open CloudPage on a 375 px wide viewport — table should show Name + Actions only, no horizontal overflow
- [ ] Resize to 480 px — Size and Type columns reappear
- [ ] Resize to desktop — all 5 columns visible, layout unchanged
- [ ] Verify dark-theme styles still apply correctly

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.scss
+++ b/frontend/src/app/pages/cloud/cloud.component.scss
@@ -145,7 +145,8 @@
   border: 1px solid var(--color-border);
   background: var(--color-surface);
   box-shadow: 0 6px 14px -14px var(--color-shadow-md);
-  overflow: hidden;
+  overflow-x: auto;
+  overflow-y: hidden;
 }
 
 :host ::ng-deep .cloud-table .p-datatable-table {
@@ -341,5 +342,33 @@
 
   .cloud-toolbar-wrap {
     top: 3.35rem;
+  }
+
+  .cloud-headline {
+    flex-wrap: wrap;
+  }
+
+  .cloud-entry-name {
+    max-width: 14rem;
+  }
+}
+
+@media (max-width: 640px) {
+  :host ::ng-deep .cloud-table .p-datatable-thead > tr > th:nth-child(2),
+  :host ::ng-deep .cloud-table .p-datatable-thead > tr > th:nth-child(3),
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(2),
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(3) {
+    display: none;
+  }
+}
+
+@media (max-width: 420px) {
+  :host ::ng-deep .cloud-table .p-datatable-thead > tr > th:nth-child(4),
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(4) {
+    display: none;
+  }
+
+  .cloud-entry-name {
+    max-width: 9rem;
   }
 }
--- a/frontend/src/app/pages/cloud/cloud.component.scss
+++ b/frontend/src/app/pages/cloud/cloud.component.scss
@@ -41,6 +41,17 @@
   padding: 0.38rem 0.48rem;
 }
 
+:host ::ng-deep .cloud-toolbar .p-toolbar-group-start,
+:host ::ng-deep .cloud-toolbar .p-toolbar-group-end {
+  min-width: 0;
+}
+
+:host ::ng-deep .cloud-toolbar .p-toolbar-group-end {
+  max-width: 100%;
+  overflow-x: auto;
+  overflow-y: hidden;
+}
+
 .cloud-toolbar-start {
   display: flex;
   align-items: center;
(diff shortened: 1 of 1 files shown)


frontend/src/app/pages/cloud/cloud.component.scss
```diff
@@ -145,7 +145,8 @@
   border: 1px solid var(--color-border);
   background: var(--color-surface);
   box-shadow: 0 6px 14px -14px var(--color-shadow-md);
-  overflow: hidden;
+  overflow-x: auto;
+  overflow-y: hidden;
```
prashantpiyush1111, line 180: Since columns are already being hidden at 640px/420px breakpoints to fit content, when exactly does horizontal scroll actually trigger? If it's a fallback for very narrow/edge-case content (e.g. long entry names not clamped by .cloud-entry-name), consider adding a short comment explaining the intent — otherwise this looks redundant with the column-hiding approach below.

frontend/src/app/pages/cloud/cloud.component.scss
```diff
@@ -342,4 +343,32 @@
...
+  }
+}
+
+@media (max-width: 640px) {
+  :host ::ng-deep .cloud-table .p-datatable-thead > tr > th:nth-child(2),
+  :host ::ng-deep .cloud-table .p-datatable-thead > tr > th:nth-child(3),
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(2),
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(3) {
+    display: none;
+  }
```
prashantpiyush1111, line 423: nth-child hardcodes column position (assumes Name=1, Size=2, Type=3, Modified=4, Actions=5). If a column is ever added/reordered (e.g. a selection checkbox), this will silently hide the wrong column. Suggest adding explicit classes (.col-size, .col-type) to the <th>/<td> templates and targeting those instead — more robust and self-documenting.

frontend/src/app/pages/cloud/cloud.component.scss
```diff
@@ -342,4 +343,32 @@
...
     top: 3.35rem;
   }
+
+  .cloud-headline {
+    flex-wrap: wrap;
+  }
+
+  .cloud-entry-name {
+    max-width: 14rem;
+  }
```
prashantpiyush1111, line 414: Minor: this changes at 768px, then again at 420px (9rem) — but there's no explicit rule between 640px and 768px accounting for the fact that Size/Type columns disappear at 640px, freeing up horizontal space. Worth double-checking the Name column doesn't look oddly narrow between 420–640px when other columns are already gone.

frontend/src/app/pages/cloud/cloud.component.scss
```diff
@@ -342,4 +343,32 @@
...
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(3) {
+    display: none;
+  }
+}
+
+@media (max-width: 420px) {
+  :host ::ng-deep .cloud-table .p-datatable-thead > tr > th:nth-child(4),
+  :host ::ng-deep .cloud-table .p-datatable-tbody > tr > td:nth-child(4) {
+    display: none;
+  }
```
prashantpiyush1111, line 430: Same fragility as above — same-column-position assumption. Also: using display: none removes this data from the accessibility tree entirely (not just visually hidden), so screen reader users lose Size/Type/Modified info completely on small screens. Is that an accepted tradeoff, or should this data be moved somewhere still accessible (e.g. an expandable row detail)?

prashantpiyush1111 (changes requested):
Test plan says "Resize to 480px — Size and Type columns reappear," but the breakpoint is max-width: 640px. At 480px, Size/Type should still be hidden, not reappear. Please clarify — is the test plan wrong, or is the breakpoint value wrong?

DenizAltunkapan (approved):

DenizAltunkapan:
@29shivam thank you!

PR #256, opened 2026-07-03
Add drag-and-drop support for external file uploads

Mahi1215:
## Summary

This PR adds support for uploading files by dragging them from the desktop into the Cloud view.

### Changes
- Added dragover and drop event handlers to the cloud table container.
- Prevented the browser's default drag behavior for external file drops.
- Reused the existing upload logic for dropped files.
- Added visual feedback while external files are dragged over the drop zone.

### Notes
- Existing drag-and-drop functionality for moving files and folders within the cloud view is preserved.

## Testing

- Verified that the frontend builds successfully.
- Reviewed that the implementation reuses the existing upload flow.
- Full end-to-end testing could not be completed because the Cloud service expected on port 8090 is not available in the local development setup.


DenizAltunkapan (approved):
@Mahi1215 thanks!

PR #266, opened 2026-07-08
feat(e2ee): implement Double Ratchet for Forward Secrecy in end-to-en…

priyanshuvishwakarma273403:
# Description

This Pull Request implements Forward Secrecy in our End-to-End Encryption (E2EE) pipeline by introducing a Double Ratchet algorithm (specifically the symmetric KDF chain ratchet). This ensures that compromised device key material cannot expose past messages.

### 1. Symmetric KDF Chain Ratchet (`e2ee.service.ts`)
- **Key Derivation (KDF)**: Added a KDF step using `crypto.subtle.deriveBits` with HKDF (SHA-256) to evolve the chain key and derive a message key for each message.
- **State Management (`RatchetState`)**: Defined a new state interface containing sending/receiving chain keys, counter states, and skipped message keys (for out-of-order messages).
- **Symmetric Initialization**: Initialized the sending and receiving chain keys deterministically from the ECDH shared secret based on device ID ordering (comparators) without needing external negotiation.
- **Payload Upgrade (v2)**: Bumped the payload version to `2` to indicate ratcheted encryption. `E2eeRecipientPayload` now includes the `counter` (sequence number) of the message.
- **Backward Compatibility**: Preserved support for v1 payload decryption. When a payload with version `1` (or undefined) is received, it gracefully falls back to the original ECDH shared secret decryption.
- **Replay Protection**: Deletes the skipped message keys immediately after use and rejects duplicate or expired keys.

### 2. Unit Testing (`e2ee.service.spec.ts`)
- Added a unit test validating **backward compatibility** by manually encrypting a v1 payload and ensuring the service decrypts it correctly.
- Added a unit test validating **out-of-order message decryption** and key skipping (e.g. decrypting message 2, skipping 0 and 1, then decrypting 0 and 1 from skipped keys).
- Added a unit test validating **replay protection** by ensuring that trying to decrypt an already-decrypted skipped key fails and returns `null`.

## Verification & QA

- **Unit tests**: Executed all frontend tests with `ng test --watch=false --browsers=ChromeHeadless` and confirmed they all passed successfully.
- **Prettier**: Formatted modified source files.
- **Production Build**: Verified that `npm run build` compiles with code 0.

## Files Modified
- `frontend/src/app/services/e2ee.service.ts`
- `frontend/src/app/services/e2ee.service.spec.ts`

### DIFF
--- a/frontend/src/app/services/e2ee.service.ts
+++ b/frontend/src/app/services/e2ee.service.ts
@@ -9,6 +9,7 @@ export interface E2eeRecipientPayload {
   iv: string;
   salt: string;
   ciphertext: string;
+  counter?: number;
 }
 
 export interface E2eePayload {
@@ -16,6 +17,14 @@ export interface E2eePayload {
   senderDeviceId: string;
   senderPublicKey: JsonWebKey;
   recipients: Record<string, E2eeRecipientPayload>;
+}
+
+interface RatchetState {
+  sendingChainKey: string;
+  receivingChainKey: string;
+  sendingCounter: number;
+  receivingCounter: number;
+  skippedMessageKeys: Record<number, string>;
 }
 
 interface StoredIdentity {
@@ -153,7 +162,7 @@ export class E2eeService {
       keyPair.publicKey,
     );
     const payload: E2eePayload = {
-      v: 1,
+      v: 2,
       senderDeviceId: deviceId,
       senderPublicKey: publicKeyJwk,
       recipients: {},
@@ -164,8 +173,10 @@ export class E2eeService {
         const recipientPublicKey = await this.importPublicKey(
           JSON.parse(device.publicKey) as JsonWebKey,
         );
-        const encrypted = await this.encryptForRecipient(
+        const encrypted = await this.encryptForRecipientWithRatchet(
           plaintext,
+          deviceId,
+          device.deviceId,
           keyPair.privateKey,
           recipientPublicKey,
           this.buildAad(device.deviceId, deviceId, publicKeyJwk),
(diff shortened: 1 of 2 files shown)


frontend/src/app/services/e2ee.service.ts
```diff
@@ -643,4 +679,257 @@ export class E2eeService {
...
+        entry.iv,
+        messageKey,
+        aad,
+      );
+      delete state.skippedMessageKeys[messageCounter];
+      this.saveRatchetState(ourDeviceId, remoteDeviceId, state);
+      return plaintext;
+    }
+
+    if (messageCounter > state.receivingCounter) {
```
DenizAltunkapan, line 759: Please add a maximum skip window before entering this loop, for example something like `MAX_SKIP = 1000`. Right now a payload with a very large `counter` can force the client to run many KDF steps and store many skipped keys, which is an easy local DoS against the browser/localStorage.

frontend/src/app/services/e2ee.service.ts
```diff
@@ -643,4 +679,257 @@ export class E2eeService {
...
+      state.receivingChainKey = this.arrayBufferToBase64(currentChainKey);
+    }
+
+    if (messageCounter === state.receivingCounter) {
+      const currentChainKey = this.base64ToArrayBuffer(state.receivingChainKey);
+      const { nextChainKey, messageKey } = await this.kdf(currentChainKey);
+
+      state.receivingChainKey = this.arrayBufferToBase64(nextChainKey);
+      state.receivingCounter++;
+      this.saveRatchetState(ourDeviceId, remoteDeviceId, state);
```
DenizAltunkapan, line 788: Please do not persist the advanced receiving state before decryption has succeeded. If the ciphertext is corrupted or malicious, this saves the new chain state anyway and the real message for this counter becomes undecryptable later. A safer flow is: derive the message key, try `decryptWithKey`, then update and save the state only after the decrypt call returns successfully.

DenizAltunkapan (changes requested):
Thanks for the work here. The direction is useful, but I would not merge this version yet because the ratchet state can be advanced in unsafe cases and a malicious or corrupted payload can force expensive local state progression. Please address the two inline points before merging.

One broader note: this is a symmetric-chain ratchet, not a full Double Ratchet with a DH ratchet step. That is fine as an incremental improvement, but the PR title/description should avoid overstating the security model unless the DH ratchet is actually implemented.

frontend/src/app/services/e2ee.service.ts
```diff
@@ -643,4 +679,268 @@ export class E2eeService {
...
+    throw new Error('Duplicate or expired message key');
+  }
+
+  private async initRatchetState(
+    ourPrivateKey: CryptoKey,
+    remotePublicKey: CryptoKey,
+    ourDeviceId: string,
+    remoteDeviceId: string,
+  ): Promise<RatchetState> {
+    const sharedSecret = await crypto.subtle.deriveBits(
```
GideonBa, line 802: This still initializes every chain from the static ECDH secret. If a device private key is compromised, an attacker can recompute the initial chain and derive message keys from the public counters, so this should not be described as Forward Secrecy or a full Double Ratchet unless a DH ratchet is added.

GideonBa (changes requested):
Thanks for addressing the previous two issues; the skip window and delayed state save look much better now. I still cannot approve this while the PR presents the change as Forward Secrecy / Double Ratchet. Please either add a real DH ratchet or update the title/description and in-code wording to describe this as a symmetric-chain ratchet only.

DenizAltunkapan:
@priyanshuvishwakarma273403 please apply the requested changes from @GideonBa

Issue #263, opened 2026-07-06
[Feature]: Show relative path tooltip for Cloud search results


DenizAltunkapan:
## Problem

Cloud search results can show multiple files with similar or hash-like names. The list currently makes it hard to understand where a search result is located without opening or interacting with it further.

Example scenario:
- A user searches inside Cloud.
- Several matching files have similar names.
- The user wants to know which folder each result belongs to before opening, editing, or downloading it.

<img width="1281" height="929" alt="Image" src="https://github.com/user-attachments/assets/b83ae481-9902-4151-96e7-0a084879dd4f" />

## Proposed solution

Show the result location when the user hovers over a search result row or filename. The tooltip should display the relative parent path of the result, using the existing search result path data if available.

This should be possible as a frontend-only change if the current search response already includes the full `path` for each result. If the backend does not provide enough path information in some cases, the issue should document the missing field before adding backend work.

## Acceptance criteria

- [ ] Search result rows or filenames expose a hover tooltip with the relative parent path.
- [ ] The tooltip uses existing `SearchResultDto.path` data where available.
- [ ] Root-level results display a clear root location, for example `/`.
- [ ] The tooltip works for both file and folder search results if both are shown.
- [ ] The UI remains usable for long paths, for example by truncating visually while preserving the full tooltip text.
- [ ] No backend changes are introduced unless the current search response lacks the required path data.

priyanshuvishwakarma273403:
hi @DenizAltunkapan  can you please assign this for me ??
thank you

DenizAltunkapan:
@priyanshuvishwakarma273403 i assigned you

PR #270, opened 2026-07-10
Feature: Show relative path tooltip for Cloud search results (#263)

priyanshuvishwakarma273403:
Problem
When performing a search inside Cloud, results can display multiple files with similar or hash-like names. Because the parent directory/path is not shown, users currently find it difficult to know which folder a search result belongs to without opening or interacting with it further.

💡 Proposed Solution
Expose a hover tooltip displaying the relative parent path of the search result. The tooltip is shown on hover over the table row when the search is active, using existing path data.

🛠️ Key Changes
Integrated TooltipModule (cloud.component.ts):

Imported PrimeNG's TooltipModule and included it in the standalone component's imports array.
Changed the visibility scope of getParentRelativePath from private to public to allow template binding.
Template Tooltip Binding (cloud.component.html):

Bound [pTooltip]="searchActive ? getParentRelativePath(entry.path) : ''" onto the table body row <tr>.
The tooltip displays only during active search. Root-level results display a clear root location: /.
Visually Safe Styling (cloud.component.scss):

Added ::ng-deep .search-path-tooltip .p-tooltip-text styling with a max-width of 24rem, word-break: break-all, and white-space: pre-wrap so that extremely long paths wrap cleanly without distorting the layout.
🧪 Verification & Quality Control
Production Build: Ran npm run build successfully to confirm no Angular template compile issues.
Linter & Formatter: Ran ESLint and Prettier, passing with 0 issues.


DenizAltunkapan:
Closing this as a duplicate. The search-path tooltip was already part of #269 (it rode along on the feat/security-activity branch by mistake), and #269 is now merged into main, so this exact change is already there. The branch names on the two PRs got swapped, which is what caused the overlap. Thanks for the work on it, no changes needed here.

PR #275, opened 2026-07-15
feat(cloud): add tooltips and aria-labels to icon-only action buttons

priyanshuvishwakarma273403:
## Summary

Improves the Cloud and Trash pages by adding `pTooltip` and `aria-label` attributes to all icon-only action buttons, completing the UI polish pass that replaced emoji icons with PrimeNG icons and ensured all HTTP errors use dismissible toast notifications.

## Problem

The Cloud page's icon-only action buttons (Download, Edit/Rename, Delete) lacked hover tooltips and screen-reader labels. Users had to guess what each icon did — especially problematic for the pencil icon which means "Edit" for files but "Rename" for folders. The same issue existed on the Trash page's Restore and Delete buttons.

## Changes

- **`cloud.component.html`** — Added `pTooltip` and `aria-label` to Download, Edit/Rename, and Delete buttons with context-aware labels (e.g. "Rename folder" vs "Edit file")
- **`trash/trash.component.html`** — Added `pTooltip` to Restore ("Restore to original location") and Delete ("Delete permanently") buttons
- **`trash/trash.component.ts`** — Imported `TooltipModule` to support the new `pTooltip` directives

## Audit Results

As part of this PR, a full codebase audit was performed:

- ✅ **Zero `alert()` calls** remain anywhere in the frontend — all HTTP errors use `UiToastService`
- ✅ **Zero emoji characters** in the Cloud/Trash pages — all action buttons use PrimeNG icons (`pi pi-trash`, `pi pi-pencil`, `pi pi-download`, `pi pi-refresh`, etc.)
- ✅ **Build passes** with no errors
- ✅ **Lint + Prettier** pass cleanly

## Testing

- [x] `npm run build` — success
- [x] `npx eslint` — no errors
- [x] `npx prettier --check` — formatted

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -212,9 +212,10 @@
                 [loading]="isDownloading(entry.path)"
                 [disabled]="isDownloading(entry.path)"
                 (click)="downloadFileByPath(entry.path, entry.name)"
-                pTooltip="Download"
+                pTooltip="Download file"
                 tooltipPosition="top"
                 ariaLabel="Download file"
+                aria-label="Download file"
               ></p-button>
               <p-button
                 icon="pi pi-pencil"
@@ -224,8 +225,13 @@
                     ? openRenameFolderDialog(entry.path, entry.name)
                     : onEditAction(entry.path, entry.name)
                 "
-                [pTooltip]="entry.kind === 'folder' ? 'Rename' : 'Edit'"
+                [pTooltip]="
+                  entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
+                "
                 tooltipPosition="top"
+                [ariaLabel]="
+                  entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
+                "
                 [attr.aria-label]="
                   entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
                 "
@@ -239,8 +245,13 @@
                     ? confirmDeleteFolder(entry.path)
                     : confirmDeleteFile(entry.path)
                 "
-                pTooltip="Delete"
+                [pTooltip]="
+                  entry.kind === 'folder' ? 'Delete folder' : 'Delete file'
+                "
                 tooltipPosition="top"
+                [ariaLabel]="
+                  entry.kind === 'folder' ? 'Delete folder' : 'Delete file'
+                "
                 [attr.aria-label]="
                   entry.kind === 'folder' ? 'Delete folder' : 'Delete file'
                 "
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -215,7 +215,6 @@
                 pTooltip="Download file"
                 tooltipPosition="top"
                 ariaLabel="Download file"
-                aria-label="Download file"
               ></p-button>
               <p-button
                 icon="pi pi-pencil"
@@ -230,9 +229,6 @@
                 "
                 tooltipPosition="top"
                 [ariaLabel]="
-                  entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
-                "
-                [attr.aria-label]="
                   entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
                 "
               ></p-button>
@@ -250,9 +246,6 @@
                 "
                 tooltipPosition="top"
                 [ariaLabel]="
-                  entry.kind === 'folder' ? 'Delete folder' : 'Delete file'
-                "
-                [attr.aria-label]="
                   entry.kind === 'folder' ? 'Delete folder' : 'Delete file'
                 "
               ></p-button>
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -212,6 +212,9 @@
                 [loading]="isDownloading(entry.path)"
                 [disabled]="isDownloading(entry.path)"
                 (click)="downloadFileByPath(entry.path, entry.name)"
+                pTooltip="Download"
+                tooltipPosition="top"
+                ariaLabel="Download file"
               ></p-button>
               <p-button
                 icon="pi pi-pencil"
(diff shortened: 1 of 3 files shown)


prashantpiyush1111 (approved):
Reviewed the changes. The implementation looks good and I didn't find any issues.

frontend/src/app/pages/cloud/cloud.component.html
```diff
@@ -212,6 +212,9 @@ <h1 class="cloud-title text-base sm:text-lg">Cloud Page</h1>
                 [loading]="isDownloading(entry.path)"
                 [disabled]="isDownloading(entry.path)"
                 (click)="downloadFileByPath(entry.path, entry.name)"
+                pTooltip="Download"
+                tooltipPosition="top"
+                ariaLabel="Download file"
```
DenizAltunkapan, line 217: This uses the PrimeNG `ariaLabel` input, while the Edit and Delete buttons just below use the native `[attr.aria-label]`, and the existing Clear-search button in this file (line 81) uses `aria-label`. Three different mechanisms for the same thing. Please make this one consistent with the others, e.g. `aria-label="Download file"`, so the rendered attribute is predictable and matches the file's convention.

frontend/src/app/pages/cloud/cloud.component.html
```diff
@@ -221,6 +224,11 @@ <h1 class="cloud-title text-base sm:text-lg">Cloud Page</h1>
                     ? openRenameFolderDialog(entry.path, entry.name)
                     : onEditAction(entry.path, entry.name)
                 "
+                [pTooltip]="entry.kind === 'folder' ? 'Rename' : 'Edit'"
```
DenizAltunkapan: Minor consistency: the tooltip says "Edit"/"Rename" but the aria-label says "Edit file"/"Rename folder", and the Delete button's tooltip is just "Delete" while its aria-label carries the file/folder context. Consider aligning tooltip and aria-label wording per button (either both short or both context-aware) so sighted and screen-reader users get the same information.

frontend/src/app/pages/cloud/trash/trash.component.html
```diff
@@ -66,6 +66,8 @@ <h1 class="trash-title text-base sm:text-lg">Trash</h1>
                 [loading]="isProcessing(entry.id)"
                 [disabled]="isProcessing(entry.id)"
                 (click)="restore(entry)"
+                pTooltip="Restore to original location"
```
DenizAltunkapan, line 69: The PR title mentions tooltips and aria-labels, but the Trash Restore button only gets a tooltip. Tooltips are not reliably exposed to screen readers, so this button still has no accessible name. Please add an `aria-label` here (e.g. "Restore to original location") to match the Cloud page.

frontend/src/app/pages/cloud/trash/trash.component.html
```diff
@@ -74,6 +76,8 @@ <h1 class="trash-title text-base sm:text-lg">Trash</h1>
                 styleClass="p-button-rounded p-button-text p-button-sm"
                 [disabled]="isProcessing(entry.id)"
                 (click)="confirmPurge(entry)"
+                pTooltip="Delete permanently"
```
DenizAltunkapan, line 79: Same as the Restore button: this permanent-delete action has a tooltip but no `aria-label`, so screen-reader users get no name for a destructive action. Please add `aria-label="Delete permanently"`.

DenizAltunkapan (changes requested):
Nice accessibility pass, and the context-aware labels (file vs folder) solve the pencil-icon ambiguity well. A few small consistency points before merge, mostly so the aria handling is uniform and the Trash buttons match what the PR title promises. All quick to address.

frontend/src/app/pages/cloud/cloud.component.html
```diff
@@ -212,6 +212,10 @@ <h1 class="cloud-title text-base sm:text-lg">Cloud Page</h1>
                 [loading]="isDownloading(entry.path)"
                 [disabled]="isDownloading(entry.path)"
                 (click)="downloadFileByPath(entry.path, entry.name)"
+                pTooltip="Download file"
+                tooltipPosition="top"
+                ariaLabel="Download file"
+                aria-label="Download file"
```
DenizAltunkapan: This button now has both the PrimeNG ariaLabel input and a native aria-label with the same value. Since p-button renders an inner button element, ariaLabel is forwarded to that real control while aria-label lands on the host wrapper, so the name ends up on two nested elements. Please keep only one mechanism. I would keep ariaLabel="Download file" and drop the aria-label line, and apply the same single-mechanism choice to the Edit and Delete buttons below so the whole file is uniform.

frontend/src/app/pages/cloud/cloud.component.html
```diff
@@ -221,6 +225,16 @@ <h1 class="cloud-title text-base sm:text-lg">Cloud Page</h1>
...
                     : onEditAction(entry.path, entry.name)
                 "
+                [pTooltip]="
+                  entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
+                "
+                tooltipPosition="top"
+                [ariaLabel]="
+                  entry.kind === 'folder' ? 'Rename folder' : 'Edit file'
+                "
+                [attr.aria-label]="
```
DenizAltunkapan: Same doubling here: this button has both [ariaLabel] and [attr.aria-label] bound to the same expression. Please drop the [attr.aria-label] and keep only [ariaLabel], so the accessible name is set once on the actual button element. The Delete button just below has the identical issue.

frontend/src/app/pages/cloud/trash/trash.component.html
```diff
@@ -66,14 +66,21 @@ <h1 class="trash-title text-base sm:text-lg">Trash</h1>
                 [loading]="isProcessing(entry.id)"
                 [disabled]="isProcessing(entry.id)"
                 (click)="restore(entry)"
+                pTooltip="Restore to original location"
+                tooltipPosition="top"
+                ariaLabel="Restore to original location"
+                aria-label="Restore to original location"
```
DenizAltunkapan: Restore and Delete here each carry both ariaLabel and aria-label with the same text. Please collapse each to a single ariaLabel so it matches the Cloud page and the name is not set twice on nested elements.

DenizAltunkapan (changes requested):
Thanks for the updates. The tooltip and aria wording is now aligned per button, which is a nice improvement. One thing still needs fixing before this can go in: several buttons now carry two aria mechanisms at once, which is the same consistency issue from the earlier review, just in a different form. Details inline. All quick to resolve.

priyanshuvishwakarma273403:
Hi @DenizAltunkapan, thanks for catching that! 

I've updated both `cloud.component.html` and `trash.component.html` to remove the redundant `aria-label` / `[attr.aria-label]` attributes. All action buttons now use PrimeNG's single `ariaLabel` / `[ariaLabel]` property uniformly so accessible names are set cleanly on the underlying button element without doubling on the host wrapper.

Pushed the fix to `feat/cloud-toast-icons`. Ready for re-review!

DenizAltunkapan (approved):
@priyanshuvishwakarma273403 thanks!

PR #280, opened 2026-07-18
fix: detect refresh token replay attacks and revoke all sessions

### LINKED ISSUE #260: Refresh token reuse (replay attack) is not detected — revoked tokens excluded from lookup queries

prashantpiyush1111:
Fixes #260

## Changes
- Added `findByTokenId(String tokenId)` to `RefreshTokenRepository` (without the revoked filter), so revoked tokens can be looked up for reuse detection.
- Updated `AuthService.refresh()`:
  - First looks up the token by `tokenId` alone.
  - If the token is found but already revoked, it's treated as a confirmed replay attack: logs a security warning and calls `revokeAllByUser()` to invalidate all other active sessions for that user.
  - Falls through to normal hash/expiry validation otherwise.

## Testing
- Verified normal login/refresh/logout flow still works.
- Verified that reusing a revoked refresh token now triggers a security log and revokes all sessions for that user, still returning a generic 401 to the client.


backend/src/main/java/vaultWeb/services/auth/AuthService.java
```diff
@@ -179,57 +98,59 @@ public ResponseEntity<?> refresh(String rawRefreshToken, HttpServletResponse res
...
+    }
+
+    if (storedToken.isRevoked()) {
+      log.warn(
+          "SECURITY_ALERT: refresh token replay detected. userId={}, tokenId={}, timestamp={}",
+          storedToken.getUser().getId(),
+          tokenId,
+          Instant.now());
+
+      return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
```
DenizAltunkapan, line 229: The PR description says this branch also calls revokeAllByUser to kill the other sessions, but the code only logs and returns 401. That session revocation is the main defense-in-depth part of the fix and issue #260 asks for it, so could you add revokeAllByUser(storedToken.getUser().getId()) here? If you would rather leave it out for now, that is fine too, but then let's update the description so it matches what the code does.

backend/src/main/java/vaultWeb/services/auth/AuthService.java
```diff
@@ -179,57 +98,59 @@ public ResponseEntity<?> refresh(String rawRefreshToken, HttpServletResponse res
     }
 
     String tokenId = claims.getId();
+    RefreshToken storedToken = refreshTokenRepository.findByTokenId(tokenId).orElse(null);
+
+    if (storedToken == null) {
+      return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
+    }
+
+    if (storedToken.isRevoked()) {
```
DenizAltunkapan: One edge case worth handling: create() revokes all tokens on every issue, so a legitimate refresh also leaves the previous token revoked. A harmless retry or double submit from the same user then lands here and gets logged as a replay alert. Rejecting it is correct, but once revokeAllByUser is added a benign retry would log the user out everywhere. Could we guard against that, for example only treating it as a replay when the incoming hash actually matches the stored one?

backend/src/main/java/vaultWeb/services/auth/AuthService.java
```diff
@@ -127,50 +87,9 @@ public User getCurrentUser() {
...
-   * <ul>
-   *   <li>{@code 401 Unauthorized} if the refresh token is missing, expired, revoked, invalid, or
-   *       reused.
-   * </ul>
-   *
-   * @param rawRefreshToken the refresh JWT provided by the client (via HttpOnly cookie)
-   * @param response HTTP response used to set the rotated refresh token cookie
-   * @return a response containing a new access token if the refresh succeeds
-   */
+  /** Refreshes the access token using a valid refresh token and performs refresh token rotation. */
```
DenizAltunkapan: This is security sensitive and there is no test covering it. Could you add one that logs in, refreshes once, then replays the old token and asserts the replay is detected? That protects the behavior from silently regressing later.

backend/src/main/java/vaultWeb/services/auth/AuthService.java
```diff
@@ -28,30 +30,20 @@
 /**
  * Service class responsible for handling authentication and user session-related operations.
  *
- * <p>Provides functionality for:
+ * <p>Provides functionality for: - Authenticating users with username and password. - Generating
```
DenizAltunkapan: Most of the deletions in this file are Javadoc being stripped or flattened into run on paragraphs, which is unrelated to the fix and makes the diff harder to read. Could you revert the doc changes and keep this PR focused on the replay detection?

DenizAltunkapan (changes requested):
Thanks for taking this on, the detection path and the generic 401 look good. A few things I would like to see before this goes in.

DenizAltunkapan:
@prashantpiyush1111 please fix the failing workflows

PR #287, opened 2026-07-20
feat(cloud): implement Secure Send sharing flow (create, list, revoke)

### LINKED ISSUE #283: [Feature]: Secure Send sharing flow in the Cloud UI (create, list, revoke)
## Problem

The Cloud Page backend now supports Secure Send links for sharing a single file with external recipients, including configurable expiry, optional password protection, listing, and revocation. None of this is available in the Vault Web frontend, so users cannot actually create or manage share links from the UI.

The existing issue Vault-Web/vault-web#281 only covers the authenticated same-origin proxy for previewing workspace files. It does not cover the Secure Send sharing flow, so that work is currently untracked.

Example scenario:
- A user wants to send one file to someone who does not have a Vault Web account.
- They want the link to expire after a set time and optionally require a password.
- Later they want to see which links are active and revoke one they no longer trust.
- Today there is no way to do any of this from the UI.

## Proposed solution

Build the Secure Send sharing flow in the Cloud UI on top of the existing backend endpoints.

Scope:
- Add a "Share link" or "Secure Send" action to the file context menu in the Cloud folder view.
- Provide a create dialog where the user sets an expiry and an optional password, then calls the create endpoint.
- After creation, show the generated URL once and let the user copy it, making clear it will not be shown again.
- Add a view that lists the current user's active links with their display name, creation date, expiry, whether a password is set, and revocation state.
- Let the user revoke a link from that list and reflect the updated state in the UI.
- Never place the password or the raw token in a query string or in logs.
- Show loading states and surface errors, including rate-limit responses, with the existing toast notifications.

Dependencies:
- The backend Secure Send endpoints already exist and were added in Vault-Web/cloud-page#104. The frontend only wires up the create, list, and revoke flows on top of them.

## Acceptance criteria

- [ ] A user can create a Secure Send link for a single file with an expiry and an optional password.
- [ ] The generated URL is shown once after creation and can be copied to the clipboard.
- [ ] A user can list their active links with expiry, password state, and revocation state.
- [ ] A user can revoke a link and see its state update.
- [ ] Passwords and raw tokens are never exposed in query strings or logs.
- [ ] Loading and error states are handled with toast notifications.
- [ ] The flow works on desktop and mobile widths.

Related: consumes the backend Secure Send endpoints from Vault-Web/cloud-page#104. This is separate from the media preview proxy tracked in Vault-Web/vault-web#281.

priyanshuvishwakarma273403:
Summary
This PR implements the end-to-end Secure Send file sharing flow in the Vault Web Cloud UI, allowing users to create expiring share links with optional password protection, view generated URLs, list active links, and revoke access on demand.

Closes #283 (Consumes backend endpoints from Vault-Web/cloud-page#104).

Problem
The Cloud backend supported Secure Send links for sharing single files with external recipients, but the Vault Web frontend lacked UI controls to trigger creation, set expiration or password protection, or list and revoke existing share links.

Proposed Solution & Key Implementation Details
DTO Models (SecureSendLinkDto.ts):

Standardized SecureSendLinkDto and CreateSecureSendRequestDto for secure data exchange.
Backend API Integration (cloud.service.ts):

Added createSecureSendLink(), listSecureSendLinks(), and revokeSecureSendLink() to connect directly with backend endpoints.
Cloud UI Integration (cloud.component.html, cloud.component.ts, cloud.component.scss):

File Action: Added a Share button (pi pi-share-alt) to file rows in the Cloud table view.
Header Action: Added a Shared Links button (pi pi-link) in the Cloud header actions.
Create Dialog: Configurable modal allowing selection of expiration periods (1 Hour, 1 Day, 7 Days, 30 Days) and optional password protection.
Generated URL Display: Modal displaying the generated share link once with a Copy to Clipboard action.
Shared Links Management View: Table modal displaying active links, creation date, expiry, password protection status, and a Revoke button that dynamically marks links as Revoked.
Security & Error Handling:

Raw passwords and tokens are never logged or exposed in query parameters.
Surface error feedback and HTTP 429 rate limits via UiToastService notifications.
Accessibility & Styling:

Uniform single ariaLabel attributes across action controls.
Styled using custom CSS badges and responsive glassmorphism themes.
Unit Tests & Quality Checks:

Unit tests added in cloud.service.spec.ts and cloud.component.spec.ts.
Verified npm run prettier:check (100% compliant) and npm run build (Exit code: 0).
Acceptance Criteria Verification
 A user can create a Secure Send link for a single file with an expiry and optional password.
 The generated URL is shown once after creation and can be copied to the clipboard.
 A user can list their active links with expiry, password state, and revocation state.
 A user can revoke a link and see its state update.
 Passwords and raw tokens are never exposed in query strings or logs.
 Loading and error states (including HTTP 429 rate limits) are handled with toast notifications.
 The flow works on desktop and mobile widths.

### DIFF
--- a/frontend/src/app/services/cloud.service.spec.ts
+++ b/frontend/src/app/services/cloud.service.spec.ts
@@ -1,0 +1,104 @@
+import { TestBed } from '@angular/core/testing';
+import {
+  HttpClientTestingModule,
+  HttpTestingController,
+} from '@angular/common/http/testing';
+import { CloudService } from './cloud.service';
+
+describe('CloudService Secure Send', () => {
+  let service: CloudService;
+  let httpMock: HttpTestingController;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [HttpClientTestingModule],
+      providers: [CloudService],
+    });
+    service = TestBed.inject(CloudService);
+    httpMock = TestBed.inject(HttpTestingController);
+  });
+
+  afterEach(() => {
+    httpMock.verify();
+  });
+
+  it('should send POST request to create a secure send link with ISO instant expiresAt', () => {
+    const mockResponse = {
+      id: 'share-123',
+      filePath: '/document.pdf',
+      fileName: 'document.pdf',
+      url: 'http://localhost/share/token123',
+      expiresAt: '2026-07-21T12:00:00Z',
+      passwordProtected: true,
+      revoked: false,
+    };
+
+    service
+      .createSecureSendLink('/document.pdf', 1440, 'secret')
+      .subscribe((res) => {
+        expect(res.id).toBe('share-123');
+        expect(res.fileName).toBe('document.pdf');
+        expect(res.hasPassword).toBeTrue();
+        expect(res.shareUrl).toBe('http://localhost/share/token123');
+      });
+
+    const req = httpMock.expectOne(`${service.apiUrl}/secure-sends`);
+    expect(req.request.method).toBe('POST');
+    expect(req.request.body.filePath).toBe('/document.pdf');
+    expect(req.request.body.expiresAt).toMatch(/^\d{4}-\d{2}-\d{2}T/);
+    expect(req.request.body.password).toBe('secret');
+    req.flush(mockResponse);
+  });
+
+  it('should map passwordProtected true from response to hasPassword true', () => {
+    const mockResponse = {
+      id: 'share-456',
+      filePath: '/secret.txt',
+      fileName: 'secret.txt',
+      url: 'http://localhost/share/token456',
+      expiresAt: '2026-07-21T12:00:00Z',
+      passwordProtected: true,
+      revoked: false,
+    };
+
+    service.listSecureSendLinks().subscribe((links) => {
+      expect(links.length).toBe(1);
+      expect(links[0].hasPassword).toBeTrue();
+      expect(links[0].isRevoked).toBeFalse();
+    });
+
+    const req = httpMock.expectOne(`${service.apiUrl}/secure-sends`);
+    expect(req.request.method).toBe('GET');
+    req.flush([mockResponse]);
+  });
+
+  it('should list active secure send links', () => {
+    const mockList = [
+      {
+        id: 'link-1',
+        filePath: '/test.png',
+        expiresAt: '2026-07-22T00:00:00Z',
+        passwordProtected: false,
+        revoked: false,
+      },
+    ];
+
+    service.listSecureSendLinks().subscribe((links) => {
+      expect(links.length).toBe(1);
+      expect(links[0].id).toBe('link-1');
+      expect(links[0].hasPassword).toBeFalse();
+    });
+
+    const req = httpMock.expectOne(`${service.apiUrl}/secure-sends`);
+    expect(req.request.method).toBe('GET');
+    req.flush(mockList);
+  });
+
+  it('should send DELETE request to revoke a secure send link', () => {
+    service.revokeSecureSendLink('link-1').subscribe();
+
+    const req = httpMock.expectOne(`${service.apiUrl}/secure-sends/link-1`);
+    expect(req.request.method).toBe('DELETE');
+    req.flush(null);
+  });
+});
--- a/frontend/src/app/services/cloud.service.ts
+++ b/frontend/src/app/services/cloud.service.ts
@@ -1,13 +1,38 @@
 import { Injectable } from '@angular/core';
 import { HttpClient, HttpParams } from '@angular/common/http';
-import { Observable } from 'rxjs';
+import { Observable, map } from 'rxjs';
 import { FolderDto } from '../models/dtos/FolderDto';
 import { FolderContentItemDto } from '../models/dtos/FolderContentItemDto';
 import { PageResponseDto } from '../models/dtos/PageResponseDto';
 import { TrashEntryDto } from '../models/dtos/TrashEntryDto';
 import { SearchResultDto } from '../models/dtos/SearchResultDto';
 import { ScanJobDto } from '../models/dtos/ScanJobDto';
+import {
+  SecureSendLinkDto,
+  CreateSecureSendRequestDto,
+} from '../models/dtos/SecureSendLinkDto';
 import { environment } from '../../environments/environment';
+
+interface RawSecureSendResponse {
+  id?: string;
+  linkId?: string;
+  token?: string;
+  shareToken?: string;
+  filePath?: string;
+  path?: string;
+  fileName?: string;
+  name?: string;
+  url?: string;
+  shareUrl?: string;
+  expiresAt?: string;
+  passwordProtected?: boolean;
+  hasPassword?: boolean;
+  protected?: boolean;
+  revoked?: boolean;
+  isRevoked?: boolean;
+  revokedAt?: string | null;
+  createdAt?: string;
+}
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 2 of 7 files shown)


GabrielBBaldez (commented):
The UI work here is solid — dialogs, revoke state, toasts and the responsive styling all look good. But the service layer never reaches the backend, so the flow can't work as it stands.

## The endpoints don't exist

`SecureSendController` from cloud-page#104 (merged) maps:

```java
@RequestMapping("/api")
@PostMapping("/secure-sends")          // plural
@GetMapping("/secure-sends")
@DeleteMapping("/secure-sends/{id}")
```

This PR calls `/secure-send` (singular), then falls back to `/shares/secure-send`, then `/shares`. All three 404 — there is no path where a link actually gets created.

The tests don't catch it because `cloud.service.spec.ts` asserts the same wrong URL (`expectOne(\`${service.apiUrl}/secure-send\`)`) and `cloud.component.spec.ts` mocks `CloudService` wholesale, so the real contract is never exercised.

## The request body doesn't match either

The backend takes an absolute instant, validated:

```java
@NotBlank private String filePath;
@NotNull @Future private Instant expiresAt;
private String password;
```

The PR sends `expiryMinutes: number`. Even at the correct URL this is a 400, since `expiresAt` would be null. The expiry dropdown can stay exactly as it is — just convert on the way out:

```ts
expiresAt: new Date(Date.now() + expiryMinutes * 60_000).toISOString()
```

## Response fields

`SecureSendDto` returns `url`, `fileName`, `createdAt`, `expiresAt`, `passwordProtected`, `revoked`.

`revoked` is read correctly. `passwordProtected` is not — the mapper looks for `hasPassword ?? protected`, so in the links table the password badge will read false for every link, including protected ones.

## Please drop the `catchError` fallback chains

Beyond the wrong URLs, chaining alternates on a **POST** is worth avoiding on principle: if the first call fails after the server already created the link, the retry creates a second one. It also defeats the rate limiter that cloud-page#104 added — a 429 on the first URL just fires two more POSTs — and the error the component finally sees comes from the *last* attempt, so a real 401 or 500 surfaces as a 404 from `/shares`. The contract is merged and known, so a single call per operation is both correct and easier to debug.

Same for the mapper's invented defaults: `id: ... || String(Date.now())` and `expiresAt: ... || now + 24h` mean a malformed response yields a link with a fabricated id (which `revoke` will then send to `DELETE`) and an expiry that doesn't match the server's. Better to let it fail loudly.

## `getStorageQuota` looks unrelated

It isn't in the PR description and doesn't seem connected to #283 — and it defaults `totalBytes` to a hardcoded `10 * 1024 * 1024 * 1024` when the backend doesn't return one, which shows the user an invented quota. Worth splitting into its own PR against whatever the real quota endpoint turns out to be.

---

Worth knowing: #274 touches `cloud.component.{ts,html,scss,spec.ts}` too, so whichever lands second will need a rebase — no action needed now, just a heads-up.

Happy to look again once the calls line up with the controller.

frontend/src/app/services/cloud.service.ts
```diff
@@ -168,4 +173,106 @@ export class CloudService {
...
+    password?: string,
+  ): Observable<SecureSendLinkDto> {
+    const normPath = this.normalizePath(filePath);
+    const payload: CreateSecureSendRequestDto = {
+      filePath: normPath,
+      expiryMinutes,
+      password: password && password.trim() ? password.trim() : undefined,
+    };
+
+    return this.http.post<any>(`${this.apiUrl}/secure-send`, payload).pipe(
```
DenizAltunkapan: The path is wrong and the request body does not match the backend. SecureSendController maps POST /api/secure-sends (plural), but this calls /secure-send (singular) and then falls back to /shares/secure-send and /shares, all of which 404, so a link is never created. Please call POST /secure-sends directly and drop the fallback chain. Chaining alternates on a POST is risky in its own right: if the first call succeeds server-side but the response is lost, the retry creates a second link, and a 429 from the rate limiter just triggers two more POSTs. A single call per operation is both correct and easier to debug. Do the same for the GET on line 216 (use /secure-sends) and the DELETE on line 233 (use /secure-sends/{id}, no /revoke and no /shares fallback).

frontend/src/app/services/cloud.service.ts
```diff
@@ -168,4 +173,106 @@ export class CloudService {
...
+
+  createSecureSendLink(
+    filePath: string,
+    expiryMinutes = 1440,
+    password?: string,
+  ): Observable<SecureSendLinkDto> {
+    const normPath = this.normalizePath(filePath);
+    const payload: CreateSecureSendRequestDto = {
+      filePath: normPath,
+      expiryMinutes,
```
DenizAltunkapan: The backend request body is { filePath, expiresAt (a @NotNull @Future Instant), password }, but this sends expiryMinutes, so even at the correct URL the create is a 400 because expiresAt is null. The expiry dropdown can stay exactly as it is; just convert on the way out, for example expiresAt: new Date(Date.now() + expiryMinutes * 60000).toISOString(), and send password only when the user set one.

frontend/src/app/services/cloud.service.ts
```diff
@@ -168,4 +173,106 @@ export class CloudService {
...
+
+    return {
+      id,
+      filePath: rawPath,
+      fileName,
+      shareUrl,
+      token,
+      expiresAt:
+        res?.expiresAt || new Date(Date.now() + 86400000).toISOString(),
+      hasPassword: Boolean(res?.hasPassword ?? res?.protected ?? !!password),
```
DenizAltunkapan: The response field is passwordProtected (see SecureSendDto), but the mapper reads hasPassword ?? protected, so the password badge in the links table will show false for every link, including protected ones. Please map from res.passwordProtected. While here, please also drop the invented defaults in this mapper: id falling back to Date.now() produces a fabricated id that revoke will then send to DELETE, and expiresAt falling back to now + 24h hides a malformed response and shows an expiry that does not match the server. Let a bad response fail loudly instead.

frontend/src/app/services/cloud.service.ts
```diff
@@ -168,4 +173,106 @@ export class CloudService {
       `${this.apiUrl}/files/trash/${encodeURIComponent(id)}`,
     );
   }
+
+  getStorageQuota(): Observable<StorageQuotaDto> {
```
DenizAltunkapan: getStorageQuota looks unrelated to this PR (it is not in the description and belongs to the quota work in #285), and it also targets endpoints that do not exist in the backend while defaulting to a hardcoded 10 GB total. Please remove it from this PR so Secure Send stays self-contained; the quota endpoint is being sorted out separately.

frontend/src/app/services/cloud.service.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    service
+      .createSecureSendLink('/document.pdf', 1440, 'secret')
+      .subscribe((res) => {
+        expect(res.id).toBe('share-123');
+        expect(res.fileName).toBe('document.pdf');
+        expect(res.hasPassword).toBeTrue();
+      });
+
+    const req = httpMock.expectOne(`${service.apiUrl}/secure-send`);
```
DenizAltunkapan: These tests assert against the same wrong URL (/secure-send) and the component spec mocks CloudService wholesale, which is why the endpoint mismatch was not caught. After switching to /secure-sends, please point the tests at the real paths, assert the outgoing body carries expiresAt as an ISO instant (not expiryMinutes), and add a case that a response with passwordProtected true maps to a protected link, so the contract is actually exercised.

DenizAltunkapan (changes requested):
The UI side of this is genuinely nice: the create dialog, the one-time URL display, the shared-links table, the revoke state and the responsive styling all look good. GabrielBBaldez already gave a thorough review and I agree with all of it, so I am not going to repeat every point. I verified the backend contract against SecureSendController and the DTOs from cloud-page#104 and confirmed the mismatches are real, so I am marking this as changes requested to make the blocking items explicit. Once the service calls line up with the merged controller this should be close to mergeable. Concrete items inline.

GabrielBBaldez (approved):
Re-checked the service against `SecureSendController` — all four blocking items are resolved: it hits `/secure-sends` (plural) on create/list/revoke with no fallback chain, sends `expiresAt` as an instant derived from the expiry dropdown, reads `passwordProtected` first in the mapper, and `getStorageQuota` is gone. The spec now asserts the real endpoint, and CI is green.

Nice work turning it around. Approving — @DenizAltunkapan this is unblocked from my side whenever you want to re-review.

DenizAltunkapan (approved):

Issue #282, opened 2026-07-19
[Feature]: Verify file integrity via checksum in the Cloud UI


DenizAltunkapan:
## Problem

The Cloud Page backend can compute a SHA-256 checksum for a stored file, but the Vault Web Cloud frontend does not expose this anywhere. Users have no way to verify that a file was uploaded or downloaded without being corrupted, and they cannot compare the stored file against a known hash.

Example scenario:
- A user uploads an important archive or document to Cloud.
- Later they want to confirm the stored copy is intact and matches the original.
- Today there is no way to see or check the file hash from the UI.

## Proposed solution

Surface the existing backend checksum endpoint in the Cloud UI as a per-file action.

Scope:
- Add a "Verify integrity" or "Show checksum" action to the file context menu in the Cloud folder view.
- Call the backend SHA-256 checksum endpoint for the selected file and display the resulting hash in a small dialog or detail panel.
- Let the user copy the hash to the clipboard.
- Optionally allow the user to paste an expected hash and show a clear match or mismatch result.
- Show a loading state while the checksum is being computed, since large files take time.
- Handle errors with the existing toast notifications rather than a generic failure.

Dependencies:
- The backend checksum endpoint already exists and was added in Vault-Web/cloud-page#95. The frontend only needs to call it and present the result.

## Acceptance criteria

- [ ] A file action in the Cloud UI returns and displays the SHA-256 checksum of the selected file.
- [ ] The displayed hash can be copied to the clipboard.
- [ ] A loading state is shown while the checksum is being computed.
- [ ] Errors are surfaced with toast notifications.
- [ ] The action and its dialog work on desktop and mobile widths.

Related: consumes the backend checksum endpoint from Vault-Web/cloud-page#95.

priyanshuvishwakarma273403:
@DenizAltunkapan  can you please assign this for me?

PR #289, opened 2026-07-21
feat(cloud): add SHA-256 file integrity checksum verification UI (#282)

priyanshuvishwakarma273403:
## 📌 Description
This PR resolves **#282** by adding a complete file integrity verification feature to the Cloud page UI. Users can now compute, view, copy, and compare the SHA-256 checksum of any stored file directly from the Cloud interface.

---

### 1. Service Layer & DTO (`CloudService` & `FileChecksumDto`)
- Created `FileChecksumDto` (`filePath`, `checksum`, `algorithm`).
- Added `getFileChecksum(filePath: string)` method in `CloudService` consuming the backend SHA-256 checksum endpoint (`GET /api/files/checksum?path=...`).
- Added flexible property mapping to gracefully handle variations in backend response formats (`checksum`, `hash`, `sha256`).

### 2. UI Action & Checksum Dialog (`CloudComponent`)
- **Per-File Action Button**: Added a dedicated *"Verify file integrity (SHA-256)"* action button (`pi pi-shield`) to every file entry row in the Cloud file browser table.
- **File Integrity Dialog (`p-dialog`)**:
  - **Header & Path**: Displays file name and full relative path.
  - **Loading State**: Shows a dedicated spinner and status text while large files are being hashed by the backend.
  - **Formatted Hash Display**: Displays computed SHA-256 hash in a monospace block for visual clarity.
  - **Clipboard Copy**: One-click copy button (`pi pi-copy`) with instant `UiToastService` feedback.
  - **Hash Verification**: Allows users to paste an expected SHA-256 hash

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -226,6 +226,15 @@
                 ariaLabel="Download file"
               ></p-button>
               <p-button
+                *ngIf="entry.kind === 'file'"
+                icon="pi pi-shield"
+                styleClass="p-button-rounded p-button-text p-button-sm cloud-icon-btn"
+                (click)="openChecksumDialog(entry)"
+                pTooltip="Verify file integrity (SHA-256)"
+                tooltipPosition="top"
+                ariaLabel="Verify file integrity (SHA-256)"
+              ></p-button>
+              <p-button
                 icon="pi pi-pencil"
                 styleClass="p-button-rounded p-button-text p-button-sm cloud-icon-btn"
                 (click)="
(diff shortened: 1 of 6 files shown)


priyanshuvishwakarma273403:
kindly review this pr.
thank you

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -1274,4 +1283,53 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+
+  copyChecksumToClipboard(): void {
+    if (!this.checksumResult?.checksum) return;
+    if (typeof navigator !== 'undefined' && navigator.clipboard?.writeText) {
+      navigator.clipboard
+        .writeText(this.checksumResult.checksum)
+        .catch(() => {});
+    }
+    this.copiedHashState = true;
+    this.toast.success(
```
DenizAltunkapan: The success toast fires unconditionally, even when the guard above skips the clipboard write (no navigator/clipboard support) or when writeText rejects. In those cases we tell the user it was copied when it was not. Move the toast into the resolved path of the promise and show an error toast on reject or when the API is unavailable.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -1274,4 +1283,53 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+      },
+    });
+  }
+
+  copyChecksumToClipboard(): void {
+    if (!this.checksumResult?.checksum) return;
+    if (typeof navigator !== 'undefined' && navigator.clipboard?.writeText) {
+      navigator.clipboard
+        .writeText(this.checksumResult.checksum)
+        .catch(() => {});
```
DenizAltunkapan: Swallowing the rejection silently hides the failure. At minimum surface it, otherwise the copied state and success toast both go out on a write that never happened.

frontend/src/app/services/cloud.service.ts
```diff
@@ -186,4 +187,35 @@ export class CloudService {
...
+                hash?: string;
+                sha256?: string;
+                value?: string;
+                algorithm?: string;
+              }
+            | string;
+          const checksum =
+            typeof raw === 'string'
+              ? raw
+              : raw?.checksum || raw?.hash || raw?.sha256 || raw?.value || '';
```
DenizAltunkapan, line 235: The response mapping accepts checksum/hash/sha256/value and defaults to an empty string on miss. If the backend contract for GET /api/files/checksum is fixed, prefer mapping the one real field and letting an unexpected shape surface as an error, rather than silently returning an empty checksum that the UI will then render and compare against. If the shape really is uncertain, add a spec covering the empty-checksum case so the dialog behaviour there is intentional.

frontend/src/app/pages/cloud/cloud.component.html
```diff
@@ -650,4 +659,135 @@ <h1 class="cloud-title text-base sm:text-lg">Cloud Page</h1>
       ></p-button>
     </ng-template>
   </p-dialog>
+
+  <!-- Checksum / File Integrity Dialog -->
+  <p-dialog
```
DenizAltunkapan, line 682: The other cloud dialogs route their dismiss through an onHide handler (see the scan dialog). This one has no (onHide), so closing via the X or Esc leaves showChecksumDialog state around without the reset the Close button path implies. Consider clearing checksumResult/expectedHash/error on hide so a reopened dialog does not briefly show stale data.

DenizAltunkapan (changes requested):
Thanks for this, the checksum flow is a nice addition and the dialog states (loading / error / result / verify) are well thought out.

Before I can approve there are a few things to sort out.

**1. Merge conflicts (blocking)**

The branch is out of date with main and conflicts in five files, all from the recent cloud work that already landed:

- frontend/src/app/pages/cloud/cloud.component.ts
- frontend/src/app/pages/cloud/cloud.component.html
- frontend/src/app/pages/cloud/cloud.component.spec.ts
- frontend/src/app/services/cloud.service.ts
- frontend/src/app/services/cloud.service.spec.ts

Please rebase onto latest main and resolve these. Since the conflicts touch the same regions as the virus-scan and unsaved-edit changes, double check nothing gets dropped on either side while resolving, and rerun the full spec suite afterwards.

The inline comments below can be handled in the same pass.

DenizAltunkapan:
@priyanshuvishwakarma273403 thanks for resolving the conflicts, but the resolution dropped tests that are on main. cloud.component.spec.ts on main has three describe blocks (virus scan, Secure Send Flow, unsaved-edit guard), your branch now only keeps the virus scan one. The Secure Send and unsaved-edit guard specs from #287 and #290 are gone, and CI is green simply because deleted tests cannot fail. Please restore those two blocks from main and merge your checksum specs alongside them.

Also, the four inline comments from my review are still unaddressed. Happy to take another look once both are in.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -1467,4 +1477,70 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+    }
+    this.copiedHashState = true;
+    this.toast.success(
```
DenizAltunkapan: This still fires the success toast even when the clipboard write was skipped (no navigator.clipboard) or when writeText rejected, so we tell the user it was copied when it was not. Move the copiedHashState flip and the success toast into the resolved branch of the promise, and show an error toast when the API is unavailable or the write fails.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -1467,4 +1477,70 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+    this.expectedHash = '';
+    this.copiedHashState = false;
+  }
+
+  copyChecksumToClipboard(): void {
+    if (!this.checksumResult?.checksum) return;
+    if (typeof navigator !== 'undefined' && navigator.clipboard?.writeText) {
+      navigator.clipboard
+        .writeText(this.checksumResult.checksum)
+        .catch(() => {});
```
DenizAltunkapan: The rejection is still swallowed here. Once the toast moves into the resolved branch, use this catch to surface a failure toast instead of an empty handler, otherwise a failed copy looks identical to a successful one.

DenizAltunkapan (changes requested):
Thanks for the follow-up, this is close now. The spec blocks from main are all restored (virus scan, Secure Send, unsaved-edit guard) alongside the new checksum specs, the conflict resolution is clean with no production code lost, and the dialog dismiss now routes through onChecksumDialogHide with a full reset. The checksumRequestId guard against stale responses is a nice extra.

One item from the earlier review is still open: the clipboard copy reports success unconditionally. Details inline. Fix that and I am happy to approve.

DenizAltunkapan (approved):

PR #290, opened 2026-07-21
feat(cloud): warn before leaving unsaved file edits

### LINKED ISSUE #261: [Feature]: Warn before leaving unsaved Cloud file edits
## Problem

When editing a file in Cloud, users can leave the editor even after changing the content. This can happen by closing the editor, navigating to another folder, opening another file, or leaving the page. If the user forgot to save, the changes are lost without a confirmation step.

Example scenario:
- A user opens a Markdown file in Cloud.
- They change the content but do not click save.
- They navigate away from the editor.
- The modified content is discarded without asking for confirmation.

## Proposed solution

Track the original file content and compare it with the current editor content. If the editor is dirty, show a confirmation dialog before allowing the user to leave or close the editor.

Suggested confirmation copy:

`Exit without saving? Your changes will be lost.`

This should apply to relevant editor exit paths, including closing the editor, opening another file, navigating to another folder, and browser/tab navigation where feasible.

## Acceptance criteria

- [ ] The Cloud file editor tracks whether the current content has unsaved changes.
- [ ] Closing the editor with unsaved changes shows a confirmation dialog.
- [ ] Navigating away from the edited file with unsaved changes requires confirmation.
- [ ] Saving the file resets the dirty state.
- [ ] Leaving an unchanged file does not show the confirmation dialog.
- [ ] Browser/tab close or refresh is guarded with a native unsaved-changes warning where supported.

GabrielBBaldez:
## What

Closes #261. The Cloud file editor could be left with unsaved changes and the edits were discarded silently — closing the dialog (the ✕, `Esc`, or a mask click), clicking **Cancel**, or leaving the page.

## How

Track the content as loaded (`originalFileContent`); the editor is *dirty* when `fileContent` diverges. Every exit path now routes through one guard, `requestCloseFileEditor()`, which asks before discarding:

> Exit without saving? Your changes will be lost.  — **Discard** / **Keep editing**

- **Cancel** → `requestCloseFileEditor()`
- **dialog dismiss** (✕ / `Esc` / mask) → `onFileEditorHide()` reverses the dismiss and routes through the same guard, so it can't sneak past the prompt
- **full-page leave** (reload, tab close, external nav) → a `window:beforeunload` guard
- a **clean** editor closes with no prompt, and **Save** is unaffected

Reuses the existing `ConfirmationService` / `<p-confirmdialog>` already used for delete confirmations — no new dependency.

## Testing

`ng test` — 5 new specs in `cloud.component.spec.ts` cover every path: clean close (no prompt), dirty Cancel (prompt, closes only on Discard), dirty dismiss (reopens + prompts), clean dismiss, and the unload guard firing only while dirty. Full suite green (22), `ng build` and Prettier clean.

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.html
+++ b/frontend/src/app/pages/cloud/cloud.component.html
@@ -367,7 +367,7 @@
 
   <p-dialog
     [(visible)]="showFileEditor"
-    (onHide)="closeFileEditor()"
+    (onHide)="onFileEditorHide()"
     [modal]="true"
     [closable]="true"
     [draggable]="false"
@@ -474,7 +474,7 @@
       <p-button
         label="Cancel"
         styleClass="p-button-text p-button-sm text-xxs sm:text-xs"
-        (click)="closeFileEditor()"
+        (click)="requestCloseFileEditor()"
       ></p-button>
       <p-button
         label="Save"
--- a/frontend/src/app/pages/cloud/cloud.component.spec.ts
+++ b/frontend/src/app/pages/cloud/cloud.component.spec.ts
@@ -409,9 +409,11 @@ describe('CloudComponent unsaved-edit guard', () => {
     );
   });
 
-  // Open the editor on a file whose loaded content is `content` (not yet dirty).
-  function openEditor(content: string): void {
+  // Open the editor on a file whose loaded state is not yet dirty.
+  function openEditor(content: string, fileName = 'note.txt'): void {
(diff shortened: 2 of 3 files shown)


GabrielBBaldez:
Heads-up @priyanshuvishwakarma273403 — saw you'd asked about this one back in July; I already had a working branch so I went ahead and opened this. Happy to hand it over or collaborate if you'd started on it.

DenizAltunkapan (approved):

Issue #277, opened 2026-07-16
[Feature]: Markdown workspace mode with split editor, live preview, outline, and wikilinks


DenizAltunkapan:
## Problem
Cloud Page will soon be able to preview Markdown files (Vault-Web/vault-web#262), but previewing is read-only. There is no way to actually write or maintain notes and documents inside Vault Web. For a private workspace this is a gap: users have to edit Markdown somewhere else and re-upload, and there is no notebook or wiki experience over the files they already store.

Example: a user keeps project notes as Markdown in their cloud. They want to open a note, edit it with live preview, jump between headings, and link to a related note, all without leaving Vault Web.

## Proposed solution
Add a Markdown workspace mode that turns Markdown files in the cloud into an editing and note-taking experience, not just a preview.

Scope:
- a split view with the editor on one side and live preview on the other
- save edits back to the cloud file, with autosave and a clear saved/unsaved indicator (reuse the unsaved-changes guard from Vault-Web/vault-web#261)
- an outline panel generated from the document headings for quick navigation
- wikilink support (for example [[other-note]]) that resolves to other Markdown files in the workspace, so notes can link to each other
- open through the shared file viewer shell (Vault-Web/cloud-page#98) rather than a separate one-off view

Notes:
- Everything stays as plain Markdown files in the existing cloud storage. No new document store, no database format, so files remain portable.
- Real-time multi-user co-editing is explicitly out of scope for this issue; this is single-user editing of shared, portable files.

## Acceptance criteria
- [ ] A Markdown file can be edited in the browser with a live preview of the rendered output.
- [ ] Edits are saved back to the same cloud file, with autosave and a visible saved/unsaved state.
- [ ] An outline of the document headings is shown and can be used to navigate.
- [ ] Wikilinks between Markdown files resolve and are navigable.
- [ ] The editor opens through the shared file viewer shell rather than a separate view.
- [ ] Leaving with unsaved changes warns the user.
- [ ] Files stay as plain Markdown in the existing storage, with no proprietary format.

Related: extends the Markdown preview (#262) and the unsaved-changes warning (#261); hosted by the workspace file viewer shell (Vault-Web/cloud-page#98).

PR #295, opened 2026-07-22
feat(cloud): add markdown workspace mode with split editor, outline, and wikilinks (#277)

priyanshuvishwakarma273403:
## Problem
While users can preview Markdown files, previewing is read-only. There is currently no way to edit, write notes, structure pages, or interlink documents inside Vault Web without downloading and re-uploading them.

## Solution
Implemented Markdown Workspace Mode in the Cloud file editor to support a premium notebook/wiki experience:

1. **Split Editor & Live Preview**: Refined layout displaying the raw editor pane alongside the sanitized live preview panel when editing Markdown files.
2. **Document Outline Navigation**: Added a responsive left sidebar pane (`markdown-outline-sidebar`) that parses all headers (`#` to `######`) and smoothly scrolls the preview to the heading when clicked.
3. **Wikilinks resolution**: Added pattern rendering `[[note-name]]` that resolves to Markdown files in the current folder, allowing clickable, inter-linked document navigation.
4. **Debounced Autosave with State Badges**: Integrates a 2-second debounced autosave mechanism with visual status badges (`Saved`, `Saving...`, `Unsaved changes`) so edits are automatically committed to the backend without manual save buttons, while preventing navigation with unsaved changes.

## Verification Details
* Added 5 comprehensive unit tests verifying outline parsing hierarchy, wikilink conversion, link click resolution, and autosave transitions.
* Ran and verified all 31 frontend specs successfully (`TOTAL: 31 SUCCESS`).
* Validated formatting with Prettier and static analysis with ESLint (0 errors).

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.ts
+++ b/frontend/src/app/pages/cloud/cloud.component.ts
@@ -90,6 +90,9 @@ export class CloudComponent implements OnInit, OnDestroy {
   fileContent = '';
   originalFileContent = '';
   originalFileName = '';
+  outline: { text: string; level: number }[] = [];
+  saveStatus: 'saved' | 'saving' | 'unsaved' = 'saved';
+  autosaveTimer: ReturnType<typeof setTimeout> | null = null;
   editorMode: 'edit' | 'preview' | 'split' = 'edit';
   previewHtml: SafeHtml = '';
 
@@ -671,6 +674,8 @@ export class CloudComponent implements OnInit, OnDestroy {
     this.fileContent = '';
     this.originalFileName = '';
     this.originalFileContent = '';
+    this.outline = [];
+    this.saveStatus = 'saved';
     this.editorMode = 'edit';
     this.previewHtml = '';
     this.showFileEditor = true;
@@ -723,8 +728,10 @@ export class CloudComponent implements OnInit, OnDestroy {
         this.originalFileName = file.name;
         this.originalFileContent = content;
         this.fileContent = content;
+        this.saveStatus = 'saved';
         this.editorMode = 'edit';
         this.updatePreview();
+        this.updateOutline();
         this.showFileEditor = true;
       },
       error: (err) => {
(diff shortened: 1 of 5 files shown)


frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -1010,13 +1226,21 @@ export class CloudComponent implements OnInit, OnDestroy {
 
   updatePreview() {
     if (!this.isMarkdownFile(this.newFileName)) return;
-    const rawMarkdown = this.fileContent || '';
+    let rawMarkdown = this.fileContent || '';
+
+    // Convert [[wikilink]] to clickable links
+    rawMarkdown = rawMarkdown.replace(
```
DenizAltunkapan: Wikilinks are turned into raw anchor HTML before marked.parse runs, and the captured target is dropped verbatim into both the data-target attribute and the link text. Anything between the double brackets is user controlled file content, so a target containing a quote or angle bracket breaks the attribute or injects markup that then goes through the parser. Escape the captured group before building the anchor, or better, do the wikilink transform after parsing on the DOM rather than string-splicing HTML into the markdown source.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -1010,13 +1226,21 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+
     try {
       // marked runs synchronously here, so the result is always a string.
       const dirtyHtml = marked.parse(rawMarkdown, { async: false });
-      // DOMPurify strips any XSS payload; only then do we tell Angular the
-      // string is safe, so the bypass never sees unsanitized HTML.
-      const cleanHtml = DOMPurify.sanitize(dirtyHtml);
+      // DOMPurify strips any XSS payload; allow data-target attributes
+      const cleanHtml = DOMPurify.sanitize(dirtyHtml, {
+        ADD_ATTR: ['data-target'],
```
DenizAltunkapan, line 1399: Allowing data-target through DOMPurify is fine, but it only makes sense together with escaping the value above. Right now the attribute value is whatever was inside the brackets, unescaped.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -993,15 +1011,213 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+      }
+
+      this.newFileName = nameToSave;
+      this.originalFileContent = this.fileContent;
+      this.originalFileName = nameToSave;
+      this.saveStatus = 'saved';
+      this.reloadCurrentFolder();
+    } catch (err: unknown) {
+      this.saveStatus = 'unsaved';
+      console.error('Autosave failed:', err);
```
DenizAltunkapan, line 1156: Autosave failures are logged to the console and the status flips back to unsaved, but the user gets no signal. Given the whole point here is to stop silent data loss, a failed autosave should surface a toast so the user knows their changes are not persisted. Also drop the console.error in favour of the existing toast service.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -993,15 +1011,213 @@ export class CloudComponent implements OnInit, OnDestroy {
...
+      this.updateOutline();
+      this.updatePreview();
+    }
+  }
+
+  triggerAutosave() {
+    if (this.autosaveTimer) {
+      clearTimeout(this.autosaveTimer);
+    }
+    this.autosaveTimer = setTimeout(() => {
```
DenizAltunkapan, line 1115: The debounce runs autosaveFile after 2s of inactivity, and autosaveFile issues a renameOrMoveFile whenever the name differs from editingFile.name. While the user is still typing a new filename that fires a server side rename on every pause, once per intermediate name. Consider only autosaving content and deferring rename to an explicit Save, or debouncing the rename separately.

frontend/src/app/pages/cloud/cloud.component.ts
```diff
@@ -993,15 +1011,213 @@ export class CloudComponent implements OnInit, OnDestroy {
...
 
+  @HostListener('window:beforeunload', ['$event'])
+  unloadNotification($event: BeforeUnloadEvent): void {
+    if (this.isEditorDirty()) {
+      $event.preventDefault();
+      $event.returnValue = 'Exit without saving? Your changes will be lost.';
+    }
+  }
+
+  canDeactivate(): Promise<boolean> | boolean {
```
DenizAltunkapan, line 1179: canDeactivate is wired into the cloud route in app.routes.ts, good. But it accepts by copying fileContent into originalFileContent before closing, which means Discard actually marks the edits as saved-state rather than reverting. Combined with the beforeunload guard from main this route now has two overlapping unsaved-change mechanisms. During the rebase, please reconcile them so there is one source of truth and Discard genuinely discards.

DenizAltunkapan (changes requested):
Solid feature and a lot of care went into the workspace UX. A few things need to happen before this can go in.

**1. Merge conflicts (blocking)**

The branch is out of date with main and conflicts in two files:

- frontend/src/app/pages/cloud/cloud.component.ts
- frontend/src/app/pages/cloud/cloud.component.scss

This is more than a textual rebase. The unsaved-edit guard from #290 already landed on main, and this branch reimplements the same thing a second way: main exposes isEditorDirty as a getter with a shared hasUnsavedEditorChanges helper and one beforeunload guard, while this PR adds isEditorDirty as a method plus its own beforeunload and canDeactivate. Resolving the conflict means picking one implementation and wiring the workspace logic on top of it, not keeping both. Please rebase onto latest main, unify the dirty tracking and the unload guard, then rerun the full spec suite.

The inline comments can be handled in the same pass.

DenizAltunkapan:
@priyanshuvishwakarma273403 nice work on the conflict resolution. The guard unification came out clean, one isEditorDirty getter, a single beforeunload, and canDeactivate now goes through closeFileEditor instead of masking the dirty state. All spec blocks from main were kept too.

Two things still open:

1. Frontend CI fails on the style budget: cloud.component.scss is 16.15 kB against a 16 kB budget, over by 152 bytes. Trimming the new workspace styles a bit should do it.

2. The other inline comments from my review are unaddressed: the wikilink target still goes unescaped into the anchor HTML before marked.parse (with ADD_ATTR letting it through), autosave failures still only log to console with no toast, and the autosave debounce still triggers a server side rename on every pause while typing a filename.

Once those are in I am happy to re-review.

DenizAltunkapan (approved):
@priyanshuvishwakarma273403 thanks

Issue #278, opened 2026-07-16
[Feature]: Service registry manifest per module for dynamic service integration


DenizAltunkapan:
## Problem
Vault Web is the portal that unites the other services (cloud, password manager, chats, habits), but it currently has to know about each service through hardcoded frontend and deployment configuration. Adding or moving a service means editing Vault Web itself: its URL, its icon, which scopes it needs, where its health endpoint is. This does not scale as the ecosystem grows and makes it easy for the portal and the actual deployment to drift apart.

Example: a new module is added to the stack. Today someone has to wire its URL and metadata into the Vault Web frontend by hand. It should be enough for the module to describe itself and for Vault Web to pick it up.

## Proposed solution
Define a small service registry manifest that each module ships, and have Vault Web build its view of the ecosystem from those manifests instead of hardcoded config.

Proposed manifest (for example vault-service.json) per module:
- name and display name
- base URL and icon
- health endpoint
- required scopes
- token forwarding expectations (how the gateway should pass identity to it)

Scope:
- agree a minimal manifest schema and document it (a good fit for server-docs)
- each module provides its manifest
- Vault Web reads the manifests and renders available services from them, rather than a hardcoded list
- unknown or unreachable services degrade gracefully in the portal

Notes:
- This should come after the gateway-centric auth work (Vault-Web/auth-api-gateway#38): the scopes and token-forwarding fields in the manifest only make sense once the gateway owns identity and audience boundaries. Sequence auth first, then dynamic registration.
- This is deployment-facing as well as portal-facing, so the deploy repo is a stakeholder even though the portal-side work lives in vault-web.

## Acceptance criteria
- [ ] A minimal service manifest schema is defined and documented.
- [ ] At least one module ships a manifest that conforms to the schema.
- [ ] Vault Web renders available services from manifests rather than a hardcoded list.
- [ ] A missing or unreachable service degrades gracefully in the portal.
- [ ] The scopes and token-forwarding fields align with the gateway auth model (#38).

Related: depends on the gateway-centric auth migration (Vault-Web/auth-api-gateway#38); the schema is a documentation candidate for server-docs.

PR #296, opened 2026-07-22
feat(portal): add dynamic service registry with manifest integration (#278)

priyanshuvishwakarma273403:
## Problem
Vault Web previously relied on a hardcoded list of services (Cloud, Password Manager) in its frontend configuration. Adding or modifying a service required editing Vault Web itself, making the portal and actual deployment prone to configuration drift.

## Solution
Introduced a dynamic service registry manifest system:

1. **Manifest Schema Definition**: Documented a standard JSON schema (`vault-service.json`) in [docs/service-manifest-schema.md](file:///d:/OpenSource/valulWeb/vault-web/docs/service-manifest-schema.md) specifying how each module describes itself (display details, route, health endpoints, required scopes, token forwarding).
2. **Mock Manifests shipped**: Shipped conformant manifests for `cloud`, `passwords`, and `habits` in the web root.
3. **Registry Startup & Health Verification**: Added `ServiceRegistryService` triggered via Angular `APP_INITIALIZER` to dynamically load manifests at boot and perform automated health checks on each service's health endpoint.
4. **Graceful UI Degradation**: Replaced hardcoded navigation links in the navbar (both desktop and mobile) with dynamically rendered items. Unreachable or missing services (such as Habits) degrade gracefully, displaying as disabled/greyed-out in the menu with an "offline" badge and descriptive tooltip.

## Verification Details
* Created unit tests in `service-registry.service.spec.ts` to assert manifest parsing, fallback handling, and health check integration.
* Verified that all unit tests pass successfully (`TOTAL: 21 SUCCESS`).
* Validated that Prettier formatting and ESLint are 100% clean (0 errors/warnings).

### DIFF
--- a/frontend/public/cloud-service.json
+++ b/frontend/public/cloud-service.json
@@ -1,0 +1,14 @@
+{
+  "name": "cloud",
+  "displayName": "Cloud",
+  "icon": "pi-cloud",
+  "route": "/cloud",
+  "baseUrl": "http://localhost:8090",
+  "healthEndpoint": "http://localhost:8090/api/health",
+  "requiredScopes": ["cloud:read", "cloud:write"],
+  "tokenForwarding": {
+    "enabled": true,
+    "type": "header",
+    "name": "Authorization"
+  }
+}
--- a/frontend/public/cloud-service.json
+++ b/frontend/public/cloud-service.json
@@ -3,8 +3,9 @@
   "displayName": "Cloud",
   "icon": "pi-cloud",
   "route": "/cloud",
-  "baseUrl": "http://localhost:8090",
-  "healthEndpoint": "http://localhost:8090/api/health",
+  "baseUrl": "http://{host}:8090",
+  "apiUrl": "http://{host}:8090/api",
+  "healthEndpoint": "http://{host}:8090/api/health",
   "requiredScopes": ["cloud:read", "cloud:write"],
   "tokenForwarding": {
     "enabled": true,
--- a/frontend/public/habits-service.json
+++ b/frontend/public/habits-service.json
@@ -2,7 +2,7 @@
   "name": "habits",
   "displayName": "Habits",
   "icon": "pi-calendar-plus",
-  "route": "http://{host}:8092",
+  "route": "/habits",
   "baseUrl": "http://{host}:8092",
   "apiUrl": "http://{host}:8092/api",
   "healthEndpoint": "http://{host}:8092/api/health",
--- a/frontend/public/habits-service.json
+++ b/frontend/public/habits-service.json
@@ -1,0 +1,14 @@
+{
+  "name": "habits",
+  "displayName": "Habits",
+  "icon": "pi-calendar-plus",
+  "route": "/habits",
+  "baseUrl": "http://localhost:8092",
+  "healthEndpoint": "http://localhost:8092/api/health",
+  "requiredScopes": ["habits:read"],
+  "tokenForwarding": {
+    "enabled": true,
+    "type": "header",
+    "name": "Authorization"
+  }
+}
--- a/frontend/public/habits-service.json
+++ b/frontend/public/habits-service.json
@@ -2,9 +2,10 @@
   "name": "habits",
   "displayName": "Habits",
   "icon": "pi-calendar-plus",
-  "route": "/habits",
-  "baseUrl": "http://localhost:8092",
-  "healthEndpoint": "http://localhost:8092/api/health",
+  "route": "http://{host}:8092",
+  "baseUrl": "http://{host}:8092",
+  "apiUrl": "http://{host}:8092/api",
+  "healthEndpoint": "http://{host}:8092/api/health",
   "requiredScopes": ["habits:read"],
   "tokenForwarding": {
     "enabled": true,
--- a/frontend/src/app/app.config.ts
+++ b/frontend/src/app/app.config.ts
@@ -1,4 +1,8 @@
-import { ApplicationConfig, provideZoneChangeDetection } from '@angular/core';
+import {
+  ApplicationConfig,
+  provideZoneChangeDetection,
+  APP_INITIALIZER,
+} from '@angular/core';
 import { provideRouter } from '@angular/router';
 import { provideHttpClient, withInterceptors } from '@angular/common/http';
 import { providePrimeNG } from 'primeng/config';
@@ -6,6 +10,17 @@ import { MessageService } from 'primeng/api';
 import Aura from '@primeuix/themes/aura';
 import { routes } from './app.routes';
 import { tokenInterceptor } from './core/interceptors/token.interceptor';
+import { ServiceRegistryService } from './services/service-registry.service';
+import { firstValueFrom } from 'rxjs';
+
+export function initializeServices(registry: ServiceRegistryService) {
+  return () =>
+    firstValueFrom(registry.loadServices())
+      .then(() => firstValueFrom(registry.checkServicesHealth()))
+      .catch((err) => {
+        console.error('Service registry initialization failed', err);
+      });
+}
 
 export const appConfig: ApplicationConfig = {
   providers: [
@@ -22,5 +37,11 @@ export const appConfig: ApplicationConfig = {
       },
     }),
     MessageService,
+    {
+      provide: APP_INITIALIZER,
+      useFactory: initializeServices,
+      deps: [ServiceRegistryService],
+      multi: true,
+    },
   ],
 };
--- a/frontend/src/app/app.config.ts
+++ b/frontend/src/app/app.config.ts
@@ -16,7 +16,9 @@ import { firstValueFrom } from 'rxjs';
 export function initializeServices(registry: ServiceRegistryService) {
   return () =>
     firstValueFrom(registry.loadServices())
-      .then(() => firstValueFrom(registry.checkServicesHealth()))
+      .then(() => {
+        registry.checkServicesHealth().subscribe();
+      })
       .catch((err) => {
         console.error('Service registry initialization failed', err);
       });
--- a/frontend/src/app/navbar/navbar.component.ts
+++ b/frontend/src/app/navbar/navbar.component.ts
@@ -9,6 +9,10 @@ import {
   ExternalDomainLink,
   resolveExternalLinkUrl,
 } from '../config/external-domains.config';
+import {
+  ServiceRegistryService,
+  ServiceManifest,
+} from '../services/service-registry.service';
 
 @Component({
   selector: 'app-navbar',
@@ -21,6 +25,7 @@ export class NavbarComponent implements OnInit {
   isMobileMenuOpen = false;
   isDomainDropdownOpen = false;
   readonly externalDomainLinks: ExternalDomainLink[] = EXTERNAL_DOMAIN_LINKS;
+  services: ServiceManifest[] = [];
 
   // Stores the full URL to the profile picture (e.g. "http://localhost:8080/uploads/...")
   // null means no picture is set — the template will show the fallback initial instead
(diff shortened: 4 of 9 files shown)


frontend/public/cloud-service.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "cloud",
+  "displayName": "Cloud",
+  "icon": "pi-cloud",
+  "route": "/cloud",
+  "baseUrl": "http://localhost:8090",
```
DenizAltunkapan: All three manifests bake in http://localhost URLs and get compiled into the frontend bundle as static assets. In any deployed environment these health checks will fail and every service renders as offline, which is the opposite of what the registry is meant to fix. The repo already derives host and protocol in environment.ts. Either serve the manifests per environment from the deployment, use relative URLs resolved against the current origin, or template the host the same way environment.ts does.

frontend/src/app/app.config.ts
```diff
@@ -1,11 +1,26 @@
...
 import Aura from '@primeuix/themes/aura';
 import { routes } from './app.routes';
 import { tokenInterceptor } from './core/interceptors/token.interceptor';
+import { ServiceRegistryService } from './services/service-registry.service';
+import { firstValueFrom } from 'rxjs';
+
+export function initializeServices(registry: ServiceRegistryService) {
+  return () =>
+    firstValueFrom(registry.loadServices())
+      .then(() => firstValueFrom(registry.checkServicesHealth()))
```
DenizAltunkapan: This chains loadServices and then all health checks inside the APP_INITIALIZER, so the whole app waits for every service to answer before anything renders, and forkJoin has no timeout. One slow or unreachable service delays startup for everyone, including the login screen which does not need any of this. Suggest awaiting only loadServices here and kicking off checkServicesHealth fire and forget, with a timeout per check.

frontend/src/app/services/service-registry.service.ts
```diff
@@ -0,0 +1,111 @@
...
+    );
+  }
+
+  checkServicesHealth(): Observable<ServiceManifest[]> {
+    if (this.services.length === 0) {
+      return of([]);
+    }
+
+    const checks = this.services.map((service) =>
+      this.http.get(service.healthEndpoint).pipe(
```
DenizAltunkapan: Two problems with this request. First, it goes through the token interceptor, which attaches the Bearer token to every non auth URL, so the auth token gets sent to whatever healthEndpoint a manifest declares. Health endpoints should not need auth, use HttpContext or a header the interceptor respects to skip token attachment. Second, when a service is down the request fails with status 0, which the interceptor treats as backend unavailable and pops the global banner at startup. A failing health check is the expected case this feature handles, it should stay silent.

frontend/src/app/navbar/navbar.component.ts
```diff
@@ -30,12 +35,14 @@ export class NavbarComponent implements OnInit {
...
     public authService: AuthService,
     private userService: UserService,
+    private serviceRegistry: ServiceRegistryService,
   ) {}
 
   /**
    * Loads the profile picture when the navbar initializes.
    */
   ngOnInit(): void {
+    this.services = this.serviceRegistry.getServices();
```
DenizAltunkapan: getServices returns a one time snapshot, so the navbar only shows anything because the blocking initializer happens to have finished first. Once the initializer stops blocking on health checks this breaks, and availability changes after boot will never reach the UI. Expose the registry state as an observable or signal and subscribe here instead.

frontend/src/app/services/service-registry.service.ts
```diff
@@ -0,0 +1,111 @@
...
+      return false;
+    }
+    const candidate = manifest as Record<string, unknown>;
+    return (
+      typeof candidate['name'] === 'string' &&
+      typeof candidate['displayName'] === 'string' &&
+      typeof candidate['icon'] === 'string' &&
+      typeof candidate['route'] === 'string' &&
+      typeof candidate['baseUrl'] === 'string' &&
+      typeof candidate['healthEndpoint'] === 'string'
```
DenizAltunkapan, line 151: apiUrl is a required field on the ServiceManifest interface but is neither checked here nor listed in the schema doc required array. Either validate it or make it optional in both places so the interface, the validator, and the doc agree.

frontend/public/habits-service.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "habits",
+  "displayName": "Habits",
+  "icon": "pi-calendar-plus",
+  "route": "/habits",
```
DenizAltunkapan: This route does not exist in app.routes.ts, so if the habits service ever reports healthy the navbar renders a link into the void. Fine to ship a mock manifest for the offline state, but worth a comment in the schema doc that route must match a registered route, or point this at a real one once habits lands.

DenizAltunkapan (changes requested):
Good direction, this is exactly what #278 asked for and the structure (manifest validation, graceful degradation, specs) is solid. No conflicts and CI is green. A few things need fixing before this can go in, mostly around deployment reality and boot behaviour.

Summary of the requested changes, details inline:

1. The manifests hardcode localhost URLs, which makes every service show as offline in any non dev deployment. That defeats the drift problem this PR sets out to solve.
2. The APP_INITIALIZER blocks app boot on all health checks with no timeout.
3. Health checks go through the token interceptor, so the auth token is sent to whatever URL a manifest declares, and a down service triggers the global backend unavailable banner at startup.
4. The navbar reads the registry once in ngOnInit, which only works because of the blocking initializer.
5. Smaller items: apiUrl is required by the interface but not validated, and the habits manifest points at a route that does not exist.

One more nit outside the diff: the PR description links to a local file path (file:///d:/...), please swap that for a repo link.

DenizAltunkapan:
Went through 47efe67c. All five points from my review are properly addressed:

- Host resolution via {host} templating and resolveUrl, including the https upgrade when the page is served over TLS
- APP_INITIALIZER no longer awaits the health checks, and both the manifest loads and the checks carry a 3s timeout
- Health checks run through a separate HttpClient built on HttpBackend, so the token interceptor and the error banners are bypassed
- Navbar is reactive now via BehaviorSubject and services$
- apiUrl is validated in isValidManifest and lines up with the schema doc

One thing to fix before I approve, and it came in with this last commit. In habits-service.json, route changed from /habits to http://{host}:8092. route is the in-app router path, not a service URL, and baseUrl already carries the host. With the new isExternalUrl branch in the navbar, habits now renders as an external anchor pointing at a bare origin instead of a router link. cloud-service.json and password-manager-service.json both kept their /cloud and /passwords paths correctly, so habits is the only one affected.

Set it back to /habits and this is good to go. The isExternalUrl handling itself is worth keeping for services that genuinely live on another origin.

prashantpiyush1111 (approved):
Update: on re-reading this against the PR's stated goal, I'm walking back my earlier approval. My review verified that the previous round's comments were resolved, but I missed that the manifest filename list is still hardcoded in loadServices() — which means the core config-drift problem this PR set out to fix isn't actually solved, just moved. Deniz's latest comments (host resolution fragility, unused isLoaded, unenforced requiredScopes/tokenForwarding, hardcoded ports, test seam) all look valid too. Agreeing these need to be addressed before merge.

frontend/src/app/services/service-registry.service.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (currentProtocol === 'https:' && resolved.startsWith('http:')) {
+        resolved = resolved.replace('http:', 'https:');
+      }
+      return resolved;
+    }
+    return url;
+  }
+
+  loadServices(): Observable<ServiceManifest[]> {
+    const manifests = [
```
DenizAltunkapan, line 64: This list of manifest filenames is still hardcoded in the app itself. The PR description says the goal is to avoid editing Vault Web when a service is added or changed, but with this array in place, adding a new service still requires a code change here. I think this needs to come from something external, like a single root manifest that lists the available services, or a build time config, otherwise we have not actually removed the config drift problem, we have just moved it from one hardcoded list to another.

frontend/src/app/services/service-registry.service.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (typeof window !== 'undefined' && window.location?.origin) {
+        return `${window.location.origin}${url}`;
+      }
+      return url;
+    }
+    if (typeof window !== 'undefined' && window.location?.hostname) {
+      const currentHost = window.location.hostname;
+      const currentProtocol = window.location.protocol;
+      let resolved = url
+        .replace('{host}', currentHost)
```
DenizAltunkapan, line 53: This string replacement approach for resolving the host is fragile. url.replace('localhost', currentHost) will replace the first occurrence of the literal text localhost anywhere in the string, not just the hostname portion, so if a path or query string ever happens to contain that word this will silently produce a broken URL. I would rather see this done with the URL API, something like constructing a URL object and only touching the hostname property, so we are not doing blind text substitution on production URLs.

frontend/src/app/services/service-registry.service.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ServiceRegistryService {
+  private servicesSubject = new BehaviorSubject<ServiceManifest[]>([]);
+  public services$ = this.servicesSubject.asObservable();
+
+  private isLoaded = false;
```
DenizAltunkapan, line 30: isLoaded is set to true in loadServices but I do not see it read anywhere else in the class. Either it is dead state that should be removed, or there was an intended guard against calling loadServices multiple times that never got wired in.

frontend/src/app/services/service-registry.service.ts
```diff
@@ -0,0 +1,154 @@
...
+export interface ServiceManifest {
+  name: string;
+  displayName: string;
+  icon: string;
+  route: string;
+  baseUrl: string;
+  apiUrl: string;
+  healthEndpoint: string;
+  requiredScopes?: string[];
+  tokenForwarding?: {
```
DenizAltunkapan, line 15: requiredScopes and tokenForwarding are defined on the manifest and documented in the schema doc, but I cannot find anywhere in this PR where they are actually read or enforced. Advertising scopes and token forwarding behavior that is not implemented yet feels risky from a security expectations standpoint, someone reading the schema doc would reasonably assume this is already handled. Could we either wire this up or leave it out of this PR until it is implemented, with a note that it is planned.

frontend/public/cloud-service.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "cloud",
+  "displayName": "Cloud",
+  "icon": "pi-cloud",
+  "route": "/cloud",
+  "baseUrl": "http://{host}:8090",
```
DenizAltunkapan, line 6: The port is still hardcoded in this manifest, same for the other two service manifests. That is probably fine for local development, but I want to understand how this is meant to work outside of localhost, for example in staging or production where services will not be sitting on fixed ports like 8090 through 8092. If that is out of scope for this PR it would help to say so explicitly, otherwise this reads as production configuration.

frontend/src/app/services/service-registry.service.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+        displayName: 'Password Manager',
+        icon: 'pi-key',
+        route: '/passwords',
+        baseUrl: service.resolveUrl('http://{host}:8091'),
+        apiUrl: service.resolveUrl('http://{host}:8091/api'),
+        healthEndpoint: pmHealthUrl,
+      },
+    ];
+
+    (service as any).servicesSubject.next(services);
```
DenizAltunkapan, line 135: Reaching into the private servicesSubject with a cast to any to set up this test is a sign the service could use a small public seam for testing, for example an optional setter or an exported testing helper. Not blocking, but worth cleaning up so the test does not depend on an internal implementation detail that could change without the test noticing.

DenizAltunkapan (changes requested):
Thanks for putting this together, the idea of moving away from a hardcoded service list is the right direction. I have a few concerns about whether this implementation actually achieves that goal, plus one thing that looks like it could cause bugs in production. Left inline notes below.

Issue #279, opened 2026-07-18
[Feature]: Download a whole folder from the Cloud UI (mobile-friendly)


DenizAltunkapan:
## Problem
The Vault Web Cloud frontend can only download files one at a time. There is no way to download a whole folder, which is especially painful on mobile where downloading many files individually is slow and awkward.

Example: a user on their phone opens a folder of photos and wants all of them at once. Today they have to download each file separately. They should be able to tap one action and get the whole folder as a single archive.

## Proposed solution
Add a "Download folder" action in the Cloud UI that downloads the current folder (or a selected folder) as a single archive, and make sure it works well on mobile.

Scope:
- a folder-level download action in the Cloud folder view and in the file viewer sidebar (Vault-Web/cloud-page#98) where appropriate
- call the backend folder-archive endpoint and trigger a normal browser download of the resulting zip
- show a clear in-progress state while the archive is being prepared or streamed, since large folders take time
- handle errors with the existing toast notifications (for example a rate-limit or a too-large folder response), not a generic failure
- ensure the action and its progress and error states are usable on small screens, matching the mobile polish already done on the Cloud page

Dependencies:
- This depends on the backend folder-archive endpoint and must not start before it lands: Vault-Web/cloud-page#105. The backend has to expose folder-as-zip download first; the frontend only wires up the action, progress, and error handling on top of it.

## Acceptance criteria
- [ ] A user can download a whole folder as a single archive from the Cloud UI.
- [ ] The action is available and usable on mobile screens.
- [ ] A clear in-progress state is shown while the archive is being prepared or downloaded.
- [ ] Errors (including rate-limit and oversized-folder responses) are surfaced with toast notifications.
- [ ] The downloaded file has the correct name and .zip extension.

Related: depends on and must follow the backend endpoint Vault-Web/cloud-page#105; the action fits into the file viewer sidebar (Vault-Web/cloud-page#98).

PR #297, opened 2026-07-22
feat(cloud): add folder download as zip archive with loading state and blob error parsing (#279)

priyanshuvishwakarma273403:
## Problem
Users could only download Cloud files individually, which was inefficient—especially on mobile devices when attempting to download an entire directory or photo album.

## Solution
Implemented a mobile-friendly folder download feature:

1. **Folder Archive Endpoint Integration**: Added `getFolderArchive(relativePath)` in `CloudService` to query `GET /folders/archive` with `responseType: 'blob'`.
2. **Dynamic UI Actions**: Added a "Download" button to the headline toolbar for downloading the current directory, as well as row-level download buttons for folder items in the main table.
3. **In-Progress Spinner States**: Reused `downloadingPaths` state to show loading spinners on active download buttons while zip archives are prepared and streamed.
4. **Blob Error Parsing**: Implemented `parseBlobError` helper using `FileReader` to extract and display detailed JSON error messages (e.g. rate limits, oversized folder warnings) surfaced by the backend instead of generic failure toasts.

## Verification
* Created unit tests in `cloud.component.spec.ts` testing zip download triggers and error blob handling.
* Verified 20/20 frontend specs pass cleanly.
* Validated ESLint and Prettier formatting (0 errors).

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.spec.ts
+++ b/frontend/src/app/pages/cloud/cloud.component.spec.ts
@@ -898,4 +898,106 @@ describe('CloudComponent unsaved-edit guard', () => {
     component.warnBeforeUnload(dirty);
     expect(dirty.preventDefault).toHaveBeenCalled();
   });
+
+  describe('folder download', () => {
+    let component: CloudComponent;
+    let cloudMock: jasmine.SpyObj<CloudService>;
+    let toastMock: jasmine.SpyObj<UiToastService>;
+
+    beforeEach(() => {
+      jasmine.clock().uninstall();
+      cloudMock = jasmine.createSpyObj<CloudService>('CloudService', [
+        'getFolderArchive',
+      ]);
+      toastMock = jasmine.createSpyObj<UiToastService>('UiToastService', [
+        'success',
+        'info',
+        'warn',
+        'error',
+      ]);
+      component = new CloudComponent(
+        cloudMock,
+        {} as never,
+        toastMock as never,
+        {} as never,
+        {} as never,
+        {} as never,
+        {} as never,
+      );
+      component.rootPath = '/root';
+      component.currentFolder = { path: '/root/sub', name: 'sub' } as never;
+    });
+
+    it('downloads the folder as a zip archive and triggers browser download', () => {
+      const mockBlob = new Blob(['zip content'], { type: 'application/zip' });
+      cloudMock.getFolderArchive.and.returnValue(of(mockBlob));
+
+      const mockAnchor = jasmine.createSpyObj<HTMLAnchorElement>(
+        'HTMLAnchorElement',
+        ['click'],
+      );
+      spyOn(document, 'createElement').and.returnValue(mockAnchor as any);
+      spyOn(URL, 'createObjectURL').and.returnValue('blob:mock-url');
+      spyOn(URL, 'revokeObjectURL');
+
+      component.downloadFolder('/root/sub', 'sub');
+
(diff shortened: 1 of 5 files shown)


frontend/src/app/services/cloud.service.ts
```diff
@@ -171,6 +171,14 @@ export class CloudService {
     });
   }
 
+  getFolderArchive(relativePath: string): Observable<Blob> {
+    const path = this.normalizePath(relativePath);
+    return this.http.get(`${this.apiUrl}/folders/archive`, {
```
DenizAltunkapan: This endpoint is not implemented in cloud-page yet (the dependency Vault-Web/cloud-page#105 is an open issue with no PR). The URL and the folderPath param name are guesses until the backend lands. Once it does, verify both against the actual controller and add rate limit and oversized folder handling if the backend returns specific statuses for those, since the acceptance criteria in #279 call them out.

frontend/src/app/pages/cloud/cloud.component.spec.ts
```diff
@@ -241,6 +241,106 @@ describe('CloudComponent virus scan', () => {
...
+        'HTMLAnchorElement',
+        ['click'],
+      );
+      spyOn(document, 'createElement').and.returnValue(mockAnchor as any);
+      spyOn(URL, 'createObjectURL').and.returnValue('blob:mock-url');
+      spyOn(URL, 'revokeObjectURL');
+
+      component.downloadFolder('/root/sub', 'sub');
+
+      expect(cloudMock.getFolderArchive).toHaveBeenCalledWith('sub');
```
DenizAltunkapan: The error specs wait on a raw setTimeout of 50ms before asserting, which is a flake risk on slow CI. parseBlobError completes its observable, so you can assert inside the subscribe callback or use fakeAsync with tick instead of a wall clock delay.

DenizAltunkapan (changes requested):
The frontend work itself is clean. It follows the existing downloadFile pattern, reuses downloadingPaths for the spinner state, the blob error parsing is a nice touch, and the row button handling both kinds is tidy. No conflicts, CI green.

One blocker though: the backend endpoint does not exist yet. Issue #279 explicitly depends on Vault-Web/cloud-page#105 and says the frontend must not start before it lands. That issue is still open with no PR, and there is no GET /folders/archive anywhere in cloud-page. As merged today, every folder download would 404. Please hold this until the backend endpoint is merged, then confirm the path and param name (folderPath) actually match what the backend exposes, since right now both are assumptions.

Two smaller inline notes below. Happy to approve once the backend dependency is in and verified against it.

DenizAltunkapan:
Update on my earlier review: the blocker I raised is resolved. The backend landed in Vault-Web/cloud-page#106 (merged 2026-07-23), which was roughly the same day I wrote that review, so my comment was already out of date when you read it. Sorry for leaving you waiting on it.

The endpoint exists now, but the contract in this PR does not match it, so folder downloads would still 404. Two things in getFolderArchive:

- Path is /api/folders/download, not /folders/archive
- Query param is path, not folderPath

So the call should target ${this.apiUrl}/folders/download with HttpParams().set('path', path). Worth noting the spec mocks CloudService, so the suite stays green either way and will not catch this.

One thing that works in your favour: that route is registered under the DOWNLOAD rate limit category in RateLimitFilter, so it really can return a JSON error body on a blob request. The blob error parsing here is the right call, not overengineering.

Everything else from my first pass still stands. The downloadFile pattern reuse, the downloadingPaths spinner state and the row button handling are all clean. Fix the URL and the param and I am happy to approve.

PR #317, opened 2026-08-18
fix(chat): derive private chat sender from authenticated security con…

### LINKED ISSUE #314: [Bug]: Unauthenticated access in PrivateChatController#getOrCreatePrivateChat allows sender impersonation
## What happened?

The `getOrCreatePrivateChat` endpoint accepts the `sender` parameter directly from the request instead of extracting it from the authenticated security context (`Authentication` principal). This allows any authenticated (or unauthenticated, depending on filter config) user to pass arbitrary `sender` values and create or view chat instances on behalf of other users.

## Steps to reproduce

1. Send a GET request to `/api/private-chats/between?sender=victimUser&receiver=otherUser`.
2. Observe that the endpoint processes the request for `victimUser` regardless of who sent the request.

## Expected

The `sender` parameter should be derived directly from `Authentication#getName()` (or the authenticated principal), preventing users from impersonating other senders.

## Actual

The endpoint accepts `sender` as a `@RequestParam`, allowing users to request or generate chats for any arbitrary user context.

## Evidence

- **Logs/error message:** Logic in `PrivateChatController.java`:
  ```java
  @GetMapping("/between")
  public PrivateChatDto getOrCreatePrivateChat(
          @RequestParam String sender,
          @RequestParam String receiver)

I would like to work on this issue.

hazelr125:
## Summary
Fixed the private chat impersonation issue by deriving the authenticated sender from the Spring Security context instead of trusting a request parameter. The endpoint now uses the current user from Authentication#getName() and only accepts the receiver from the request.

## Linked issue
Closes #314

## How to test
a. Authenticate as a user and call /api/private-chats/between?receiver=otherUser
b. Confirm the chat is created/retrieved for the authenticated user only
c. Try passing a different sender value in the request and verify it is ignored
d. Confirm the endpoint still works normally for valid authenticated requests

## Notes / Risk
a. Low risk: this is a controller-level authorization fix
b. Frontend callers should stop sending sender in the query string
c. No database migration or config change required

### DIFF
--- a/backend/src/main/java/vaultWeb/controllers/PrivateChatController.java
+++ b/backend/src/main/java/vaultWeb/controllers/PrivateChatController.java
@@ -57,9 +57,9 @@ public class PrivateChatController {
       responseCode = "401",
       description = "Unauthorized request. You must provide an authentication token.")
   public PrivateChatDto getOrCreatePrivateChat(
-        @RequestParam String receiver, Authentication authentication) {
+    @RequestParam String receiver, Authentication authentication) {
     if (authentication == null) {
-        throw new UnauthorizedException("User not authenticated");
+    throw new UnauthorizedException("User not authenticated");
     }
     String sender = authentication.getName();
     PrivateChat chat = privateChatService.getOrCreatePrivateChat(sender, receiver);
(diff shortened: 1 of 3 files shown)


DenizAltunkapan (approved):
@hazelr125 thanks!

PR #318, opened 2026-08-18
Fix duplicate message shown to sender in group and private chats (#298)

### LINKED ISSUE #298: Group chat message is shown twice for the sender
## Summary

Sending a message in a group chat shows it twice in the sender's own view. The message is stored and delivered only once, the duplicate is a client side rendering problem.

## Cause

After a successful publish, the composer appends the message locally so the sender sees it immediately:

    this.decryptAndAppendMessage(message)

The backend then broadcasts the saved message to /topic/group/{id}, and the sender is subscribed to that topic as well, so the same message arrives a second time. The guard that is supposed to catch this is isDuplicateMessage in private-chat-dialog.component.ts:

    existing.privateChatId === message.privateChatId &&
    existing.groupId === message.groupId &&
    existing.senderUsername === message.senderUsername &&
    existing.timestamp === message.timestamp

For a group message the locally appended view holds privateChatId as undefined, because toViewMessage normalises falsy ids with a logical or to undefined. The DTO echoed by the server carries privateChatId as null. The strict comparison of undefined against null fails, the guard reports "not a duplicate", and the message is appended a second time.

The same mismatch exists in mirror image for private chats, where the local view holds groupId as undefined while the echo carries null.

## Evidence

The comparison is present in the deployed bundle in exactly this form:

    privateChatId===n.privateChatId&&e.groupId===n.groupId&&e.senderUsername===n.senderUsername&&e.timestamp===n.timestamp

## Expected behaviour

A message that the sender has already appended locally is not appended again when the server echo for the same message arrives.

## Suggested fix

Normalise the conversation ids before comparing, for example by comparing with a null coalescing step on both sides, so that null and undefined are treated as equal. Comparing the timestamps as parsed instants instead of as strings would additionally protect against a formatting difference: the client sends an ISO string from Date.toISOString, which always carries three millisecond digits, while the server returns Instant.toString, which drops the millisecond part when it is zero.

A more durable alternative is to give every outgoing message a client generated id that is stored and echoed back, and to deduplicate on that id.

AnUbH75:
## Summary
Sending a message in a group chat showed it twice in the sender's own view — the message was stored and delivered only once, the duplicate was a client-side rendering issue.

`isDuplicateMessage` in `private-chat-dialog.component.ts` compared `privateChatId`/`groupId` with strict equality. The locally-appended view held `privateChatId` as `undefined` (normalized by `toViewMessage`), while the server-echoed DTO carried `privateChatId: null`. Since `undefined === null` is `false`, the guard failed to detect the echo as a duplicate. The mirror-image issue existed for private chats with `groupId`.

Two changes fix this:
1. Normalize ids with nullish coalescing (`?? null`) before comparing, and compare timestamps as parsed instants rather than raw strings (defends against the `Date.toISOString()` vs `Instant.toString()` formatting difference).
2. Added a client-generated `clientMessageId` (UUID), attached on send, persisted, and echoed back unchanged — a more durable dedup key than field-matching. Falls back to the normalized comparison for messages without one (backward compatible, nullable column).

## Linked issue
Closes #298

## How to test
1. Start the stack (`docker compose up -d`, backend, frontend).
2. Open two browser sessions as different users, join/create a group chat with both.
3. Send a message from User A — confirm it appears **once** for A and once for B.
4. Repeat in a private (1:1) chat to cover the mirror-image `groupId` case.
5. Unit tests:
   - Backend: `./mvnw test -Dtest=ChatServiceTest`
   - Frontend: `ng test --include='**/private-chat-dialog.component.spec.ts'`

## Notes / Risk
- **No migration needed** — `spring.jpa.hibernate.ddl-auto=update` auto-applies the new nullable `client_message_id` column on `chat_message`.
- **Backward compatible** — existing messages without a `clientMessageId` fall back to the original (now-fixed) field comparison, so no data backfill required.
- Low risk: additive schema change, no removed/renamed fields, no breaking API changes.

### DIFF
--- a/backend/src/main/java/vaultWeb/dtos/ChatMessageDto.java
+++ b/backend/src/main/java/vaultWeb/dtos/ChatMessageDto.java
@@ -21,4 +21,6 @@ public class ChatMessageDto {
   private String e2eePayload;
   private MessageType messageType;
   private PollResponseDto poll;
+
+  private String clientMessageId;
 }
--- a/backend/src/main/java/vaultWeb/models/ChatMessage.java
+++ b/backend/src/main/java/vaultWeb/models/ChatMessage.java
@@ -43,4 +43,7 @@ public class ChatMessage {
   @OneToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "poll_id")
   private Poll poll;
+
+  @Column(name = "client_message_id", length = 36, nullable = true)
+  private String clientMessageId;
 }
--- a/backend/src/main/java/vaultWeb/services/ChatService.java
+++ b/backend/src/main/java/vaultWeb/services/ChatService.java
@@ -81,6 +81,7 @@ public class ChatService {
     ChatMessage message = new ChatMessage();
     message.setSender(sender);
     message.setMessageType(messageType);
+    message.setClientMessageId(dto.getClientMessageId());
 
     if (dto.getTimestamp() != null) {
       message.setTimestamp(Instant.parse(dto.getTimestamp()));
@@ -136,6 +137,8 @@ public class ChatService {
     dto.setSenderUsername(message.getSender().getUsername());
 
     dto.setSenderDeviceId(message.getSenderDeviceId());
+
+    dto.setClientMessageId(message.getClientMessageId());
 
     MessageType messageType =
         message.getMessageType() != null ? message.getMessageType() : MessageType.TEXT;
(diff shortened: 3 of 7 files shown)


DenizAltunkapan (approved):
@AnUbH75 thanks!

AnUbH75:
> @AnUbH75 thanks!

Thank you for the opportunity will contribute more to this repo @DenizAltunkapan

PR #319, opened 2026-08-19
fix(frontend): add scroll area to security activity timeline

### LINKED ISSUE #301: Security activity page renders all events without a scroll area
## Summary

The security activity page renders every event it receives in one long, unbounded list. The page grows very tall and there is no way to scan the recent entries without scrolling the whole document.

## Current behaviour

The backend already caps the result set, the repository query is findTop50ByUserOrderByTimestampDesc, so up to fifty events are returned. The component assigns the whole array and the template iterates over all of it. The stylesheet defines no maximum height and no overflow for the timeline area, so there is no inner scroll region either.

## Expected behaviour

Either the list area has its own scroll container with a sensible maximum height, or only the most recent entries are shown with an option to load more. Both are acceptable, the second scales better if the server side cap is ever raised.

## Suggested fix

The smallest change is a maximum height plus vertical overflow on the timeline wrapper, so the surrounding page stays short. If a load more control is preferred, the component can render a slice of the events and grow it on demand, and the endpoint can later take a limit or an offset parameter.

The counter badge should keep showing the total number of loaded events, not the number of currently visible ones.

hazelr125:
## Summary
Added a maximum height and vertical scrolling to the security activity timeline. The page remains compact while all loaded events remain accessible.

## Linked issue

Closes #301 

## How to test
a. Open the Security Activity page.
b. Ensure more events exist than fit in the timeline area.
c. Verify the timeline scrolls independently.
d. Verify the badge still shows the total number of loaded events.

## Notes / Risk
CSS-only change. No migrations, feature flags, or rollout changes.

### DIFF
--- a/frontend/src/app/pages/security-activity/security-activity.component.scss
+++ b/frontend/src/app/pages/security-activity/security-activity.component.scss
@@ -68,6 +68,10 @@
 .timeline-wrap {
   position: relative;
   padding-left: 2rem;
+  max-height: min(65vh, 720px);
+  overflow-y: auto;
+  padding-right: 0.75rem;
+  scrollbar-gutter: stable;
 }
 
 .timeline {


DenizAltunkapan (approved):
thanks!

PR #320, opened 2026-08-19
Fix editor reloading folder while typing (#300)

### LINKED ISSUE #300: Remove editor autosave, it reloads the current folder while typing
## Summary

While editing a file in the cloud editor, the content is saved automatically two seconds after the last keystroke, and every autosave reloads the current folder. The file list flickers and reloads while the user is still typing. Automatic saving should be removed, saving stays an explicit action.

## Cause

onContentChange schedules triggerAutosave with a two second debounce. autosaveFile uploads the content and then calls reloadCurrentFolder.

## Evidence

The sequence is present in the deployed bundle:

    this.originalFileContent=this.fileContent,this.saveStatus="saved",this.reloadCurrentFolder()

directly followed by the autosave error branch.

## Expected behaviour

Editing content never writes to the server on its own. The file is written when the user saves explicitly. The unsaved indicator keeps showing that there are pending changes, and the existing confirmation prompt when closing an editor with unsaved changes stays in place.

## Suggested fix

Remove the autosave timer and the autosave method, and keep the dirty state tracking that drives the status label and the close confirmation. The Markdown outline and the preview update stay bound to the content change handler, since those are local and do not touch the server.

Please keep the behaviour when creating a new Markdown file intact, the preview there currently works as expected.

AnUbH75:
## Summary
Removes the cloud editor's autosave timer, which was silently reloading the current folder every ~2s of typing, causing the file list/table to flicker while the user is still editing.

Root cause: `autosaveFile()` uploaded the file content in the background and then called `reloadCurrentFolder()` on success. This ran on a 2-second debounce (`triggerAutosave()`) from every `onContentChange()` keystroke, so any sustained typing triggered repeated background folder reloads.

Fix follows the approach suggested in the issue: remove the autosave timer and `autosaveFile()`/`triggerAutosave()` methods entirely, rather than just dropping the reload call. Saving now happens only through the existing explicit Save action (`saveFile()`), which is untouched.

Preserved, unchanged:
- Dirty-state tracking (`saveStatus`, `isEditorDirty`) — still updates live as the user types
- Close-confirmation guard (`requestCloseFileEditor`, `onFileEditorHide`, `canDeactivate`, `warnBeforeUnload`) — still prompts on any unsaved exit path
- Local Markdown outline/preview updates (`updateOutline()`, `updatePreview()`) — still update on every keystroke via `onContentChange()`, since these don't touch the server
- New-file creation and preview behavior

## Linked issue
Closes #300

## How to test
1. Open the Cloud editor on an existing (or new) Markdown/text file.
2. Type continuously for more than 2 seconds.
3. **Before this fix:** the folder table below/behind the editor would flicker/reload mid-typing.
4. **After this fix:** no reload occurs while typing; the "unsaved" status indicator still updates correctly, and the outline/preview panels still update live.
5. Click **Save** — content is written and the folder list refreshes as expected (unchanged explicit-save behavior).
6. Try closing the editor with unsaved changes (Cancel button, dialog dismiss, and a full page refresh/tab close) — confirmation prompt still appears in all three paths.

Unit tests: `cloud.component.spec.ts` → `CloudComponent Unsaved Changes Flow > Markdown Workspace Mode`
- `does not autosave or reload the folder while typing (#300 regression)` — asserts no `uploadFile`/`getFolderByPath` call after 5s of simulated typing, and that `saveStatus` stays `'unsaved'`.
- `still updates the local outline and preview while typing, without saving` — asserts outline/preview still update locally with no network calls.

## Notes / Risk
- Low risk: this only removes a background-save code path; the explicit Save flow (`saveFile()`) is untouched.
- Behavior change: files are no longer saved automatically in the background. Users must click Save (or will be prompted on exit) to persist changes. This matches the fix suggested in the linked issue.
- Unrelated to this change: there's a pre-existing test failure in `CloudComponent unsaved-edit guard > folder download > downloads the folder as a zip archive...`, caused by `jasmine.createSpyObj` returning a non-DOM object passed to `document.body.appendChild`. Not touched by this PR; flagged separately.

### DIFF
--- a/frontend/src/app/pages/cloud/cloud.component.spec.ts
+++ b/frontend/src/app/pages/cloud/cloud.component.spec.ts
@@ -490,7 +490,7 @@ describe('CloudComponent Unsaved Changes Flow', () => {
       });
     });
 
-    it('should transition saveStatus through unsaved -> saving -> saved during autosave', async () => {
+    it('does not autosave or reload the folder while typing (#300 regression)', () => {
       component.showFileEditor = true;
       component.newFileName = 'note.md';
       component.fileContent = 'new markdown';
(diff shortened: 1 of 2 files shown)


AnUbH75:
@DenizAltunkapan hi can you please validate this PR

DenizAltunkapan (approved):
@AnUbH75 sorry, i was busy. thanks for the contribution!

PR #324, opened 2026-08-24
Feature issue#111

AnUbH75:
## Summary

Adds the frontend UI for the two backend behaviors added in https://github.com/Vault-Web/cloud-page/issues/111: creating a link with no expiry ("Never"), and permanently deleting a link, distinct from the existing revoke.

- `expiryOptions` on the create-share dialog gains a "Never" option. `createSecureSendLink()` now accepts `expiryMinutes: number | null`, passing `expiresAt: null` straight through when "Never" is selected instead of computing a future ISO timestamp.
- `SecureSendLinkDto.expiresAt` is now `string | Date | null`, and `mapSecureSendLink()` preserves a genuinely missing `expiresAt` as `null` (previously coerced to `''`, which broke expiry checks against a never-expiring link).
- New `lastAccessedAt` field on `SecureSendLinkDto`, surfaced as a "Last opened" column on the Public links table, so an owner can spot stale links now that links can live indefinitely.
- `linkStatus()` gains a `never-expires` state (styled distinctly, blue) alongside the existing `active`/`expired`/`revoked` states; `isExpired()` is null-safe so a never-expiring link no longer reads as expired.
- New "Delete" button next to the existing "Revoke" button on each public link row, calling a new `deleteSecureSendLink()` → `DELETE /secure-sends/{id}/permanent`. Unlike Revoke, Delete is enabled regardless of link status, since its purpose is tidying up stale/dead links (including already-revoked or expired ones). The existing revoke flow and its endpoint are untouched.

## Linked issue

Closes #111

## How to test

1. Open the create-share dialog on any file, select "Never" from the expiry dropdown, and create the link.
2. Go to Sharing → Public links: confirm the row shows "Never" in the Expires column and "Never opened" under Last opened.
3. Open the generated link once, then hit Refresh on the Sharing page: confirm Last opened now shows a real timestamp.
4. Click Delete on that link: confirm it disappears from the list immediately, and that opening the same link afterward fails (unavailable), not just after the nightly cleanup job runs.
5. Confirm Revoke still behaves exactly as before on a link with a normal (non-null) expiry — no regression.

## Notes / Risk

- Depends on the backend changes in https://github.com/Vault-Web/cloud-page/pull/120 (nullable `expiresAt`, `lastAccessedAt` tracking, and the new `DELETE /secure-sends/{id}/permanent` endpoint) — needs that PR merged/deployed first, or this UI will call an endpoint that doesn't exist yet.
- No new dependencies. Purely additive to the existing Sharing page — no changes to unrelated flows (member shares, "Shared with me").

### DIFF
--- a/frontend/src/app/models/dtos/SecureSendLinkDto.ts
+++ b/frontend/src/app/models/dtos/SecureSendLinkDto.ts
@@ -8,11 +8,14 @@ export interface SecureSendLinkDto {
   resourceType?: SharedResourceType;
   shareUrl?: string;
   token?: string;
-  expiresAt: string | Date;
+  /** Null means the link never expires. */
+  expiresAt: string | Date | null;
   hasPassword?: boolean;
   isRevoked?: boolean;
   revokedAt?: string | Date | null;
   createdAt: string | Date;
+  /** Null if the link has never been opened. */
+  lastAccessedAt?: string | Date | null;
 }
 
 /**
@@ -29,6 +32,7 @@ export interface PublicShareDto {
 
 export interface CreateSecureSendRequestDto {
   filePath: string;
-  expiresAt: string;
+  /** Omit or null for a link that never expires. */
+  expiresAt: string | null;
   password?: string;
 }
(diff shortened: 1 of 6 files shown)


AnUbH75:
@DenizAltunkapan please review this PR this is frontend changes for [issue#111](https://github.com/Vault-Web/cloud-page/issues/111) in cloud page

DenizAltunkapan (approved):

PR #325, opened 2026-08-26
Feature issue#292

### LINKED ISSUE #292: [Feature]: Add a built-in customizable Password Generator in Password Manager
### Problem
Currently, when creating or editing a password entry in the Password Manager, users have to manually type or copy a password generated from an external tool. There is no helper to generate strong, secure, and random passwords within the application.

### Proposed Solution
Add an inline password generator component or modal directly inside the "Create/Edit Password" form. Next to the password input field, a "Generate" button should open a customizable generator widget.

### Key Requirements (Acceptance Criteria)
- An inline button next to the password input that toggles the generator utility.
- Customization options:
  - **Length slider** (default 16, range 8 - 64 characters).
  - **Character toggles**: Uppercase, Lowercase, Numbers, and Special Characters.
  - **Readability toggle**: Option to generate easy-to-read characters (avoiding confusing letters like `l, 1, O, 0`).
- A "Copy to Input" button that automatically fills both the `Password` and `Confirm Password` fields in the form.
- Visual strength indicator (e.g., Weak, Medium, Strong) as the password updates.

### Technical Implementation Notes
- This can be implemented entirely client-side in the frontend without requiring backend changes, utilizing safe random generation libraries (like browser `crypto.getRandomValues`).
- Reuses PrimeNG inputs and sliders to keep styling consistent with the dashboard's design system.

AnUbH75:
## Summary

Implemented the built-in customizable password generator for the Password Manager.

* Added an inline **Generate** control next to the password field.
* Added configurable password length from **8–64 characters**, with a default length of 16.
* Added character-set options for uppercase, lowercase, numbers, and special characters.
* Added an option to avoid ambiguous characters such as `l`, `I`, `1`, `O`, and `0`.
* Added password strength feedback with **Weak / Medium / Strong** indicators.
* Added the ability to generate a new password and apply it directly to the Password and Confirm Password fields.
* Integrated the generator directly into the existing Password Manager form and styling.
* Kept the implementation client-side, with no backend or database changes.

This addresses the requirements described in issue #292.

## Linked issue

Closes #292

## How to test

1. Open the Password Manager and click **Add password**.
2. Open the **Generate** control beside the Password field.
3. Verify that the generator opens inline.
4. Change the password length using the slider and verify the generated password updates.
5. Toggle uppercase, lowercase, numbers, special characters, and the ambiguous-character option.
6. Click **New password** or the regenerate icon and verify a new password is generated.
7. Verify the strength indicator changes according to the generated password.
8. Click **Use this password** and verify that both **Password** and **Confirm Password** are populated with the generated password.
9. Repeat the flow while editing an existing password.
10. Verify the generator and password form remain usable on smaller screen sizes.

## Notes / Risk

No backend, database, migration, feature flag, or API changes are required. The password generator is implemented on the client side and integrated into the existing Password Manager form.

The changes are scoped to the Password Manager/password generator functionality.


AnUbH75:
Hi @DenizAltunkapan please review this MR

frontend/src/app/pages/password-manager/password-manager.component.scss
```diff
@@ -84,320 +131,839 @@
...
+
   border-radius: 999px;
-  background: var(--color-primary);
-  color: #fff;
+
+  background: var(--color-border);
+}
+
+.strength-bar-fill {
+  width: 100%;
```
prashantpiyush1111: The strength bar is always rendered at 100% width, so weak/medium/strong only changes the color and not the visual strength level. Could we use different widths for each strength (e.g. 33%, 66%, 100%) or bind the width dynamically so the bar actually represents the calculated strength?

prashantpiyush1111 (commented):
Could the bar width reflect the password strength instead of always being 100%?

frontend/src/app/pages/password-manager/password-manager.component.scss
```diff
@@ -84,320 +131,839 @@
...
+
+  box-shadow: var(--pm-shadow);
+}
+
+.section-header {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
+  gap: 1rem;
+
```
DenizAltunkapan, line 541: This confirms the earlier comment about the bar always showing full width. strength-bar-fill has a fixed width of 100 percent here, and the weak, medium, strong classes below only change background color, not width. So the bar visually looks the same length no matter what generatorStrength returns, only the color changes. A simple fix is to bind width or a transform scaleX to the strength value from the component, for example 33 percent for weak, 66 for medium, 100 for strong.

DenizAltunkapan (changes requested):
Checked the crypto.getRandomValues generation logic closely, that part is solid and uses correct rejection sampling to avoid modulo bias. Confirming the open comment about the strength bar below with the actual root cause.

AnUbH75:
Hi @prashantpiyush1111 thank you for reviewing. The strength bar was always rendered at 100%, so only the color reflected the password strength. I have updated it so the fill width also reflects the calculated strength. 

please check @DenizAltunkapan, @prashantpiyush1111

prashantpiyush1111 (approved):
LGTM — did two independent verification passes. Confirmed the crypto.getRandomValues rejection-sampling logic correctly avoids modulo bias, verified all four new generator buttons have type="button" (no accidental form-submit risk), and checked the new .input-actions CSS scoping against the existing .input-group > .icon-btn-inline rule used on the Confirm Password field — no conflict. Strength-bar width fix from earlier review also looks correct. No concerns, approving.

DenizAltunkapan (approved):
Confirmed the fix, strength-bar-fill now uses width 33/66/100 percent tied to the weak/medium/strong classes instead of a fixed 100 percent. The generation logic (crypto.getRandomValues with proper rejection sampling to avoid modulo bias) was already solid before and is unchanged. Approving.

PR #329, opened 2026-08-28
feat(chat): add rich-text Markdown support

### LINKED ISSUE #294: [Feature]: Rich-Text Markdown support in Chat messages
### Problem
Currently, chat messages (both private and group chats) are rendered as plain text. This limits collaboration because developers cannot share structured list items, bold/italic text emphasis, or formatted code snippets/code blocks.

### Proposed Solution
Render chat messages using Markdown syntax. Since `marked` and `dompurify` are already included as dependencies in the frontend's `package.json` (and used in the Cloud markdown editor), we can leverage them to parse chat messages dynamically.

### Key Requirements (Acceptance Criteria)
- Parse incoming chat message text using `marked` parser.
- Sanitize the output HTML using `DOMPurify` to block XSS payloads.
- Apply clean CSS styling for code blocks (using monospaced font), lists, blockquotes, and inline formatting inside the chat bubble.
- Add code block syntax highlighting support (optional, can be done via CSS).

### Technical Implementation Notes
- Implement a custom Angular pipe (e.g., `MarkdownPipe`) that takes the raw text message, compiles it to HTML, sanitizes it, and returns it as `SafeHtml`.
- Use this pipe in both private chat dialogs and group chat components.

prashantpiyush1111:
## Summary
- render private and group chat messages as Markdown using `marked`
- sanitize rendered HTML with `DOMPurify` after parsing
- add styling for lists, blockquotes, inline code, and fenced code blocks
- add tests covering Markdown rendering and XSS sanitization

Closes #294

### DIFF
--- a/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
+++ b/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
@@ -34,6 +34,10 @@ export function renderChatMarkdown(content: string): string {
     breaks: true,
     async: false,
   }) as string;
-  const cleanHtml = DOMPurify.sanitize(rendered);
+  const withSafeLinkAttributes = rendered.replace(
+    /<a\b(?![^>]*\btarget=)([^>]*)>/gi,
+    '<a target="_blank" rel="noopener noreferrer"$1>',
+  );
+  const cleanHtml = DOMPurify.sanitize(withSafeLinkAttributes);
   return `<div class="chat-markdown">${cleanHtml}</div>`;
 }
--- a/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
+++ b/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
@@ -30,16 +30,10 @@ function installStyles(): void {
 
 export function renderChatMarkdown(content: string): string {
   installStyles();
-  const rendered = marked.parse(content ?? '', { async: false });
+  const rendered = marked.parse(content ?? '', {
+    breaks: true,
+    async: false,
+  }) as string;
   const cleanHtml = DOMPurify.sanitize(rendered);
   return `<div class="chat-markdown">${cleanHtml}</div>`;
 }
-
-void import('./private-chat-dialog.component').then(
-  ({ PrivateChatDialogComponent }) => {
-    const prototype = PrivateChatDialogComponent.prototype as unknown as {
-      formatMessage: (content?: string) => string;
-    };
-    prototype.formatMessage = (content = '') => renderChatMarkdown(content);
-  },
-);
--- a/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
+++ b/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
@@ -35,9 +35,11 @@ export function renderChatMarkdown(content: string): string {
   return `<div class="chat-markdown">${cleanHtml}</div>`;
 }
 
-void import('./private-chat-dialog.component').then(({ PrivateChatDialogComponent }) => {
-  const prototype = PrivateChatDialogComponent.prototype as unknown as {
-    formatMessage: (content?: string) => string;
-  };
-  prototype.formatMessage = (content = '') => renderChatMarkdown(content);
-});
+void import('./private-chat-dialog.component').then(
+  ({ PrivateChatDialogComponent }) => {
+    const prototype = PrivateChatDialogComponent.prototype as unknown as {
+      formatMessage: (content?: string) => string;
+    };
+    prototype.formatMessage = (content = '') => renderChatMarkdown(content);
+  },
+);
--- a/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
+++ b/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
@@ -30,7 +30,7 @@ function installStyles(): void {
 
 export function renderChatMarkdown(content: string): string {
   installStyles();
-  const rendered = marked.parse(content ?? '');
+  const rendered = marked.parse(content ?? '', { async: false });
   const cleanHtml = DOMPurify.sanitize(rendered);
   return `<div class="chat-markdown">${cleanHtml}</div>`;
 }
--- a/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
+++ b/frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
@@ -1,0 +1,43 @@
+import DOMPurify from 'dompurify';
+import { marked } from 'marked';
+
+const MARKDOWN_STYLES = `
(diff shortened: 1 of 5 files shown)


frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function renderChatMarkdown(content: string): string {
+  installStyles();
+  const rendered = marked.parse(content ?? '', { async: false });
+  const cleanHtml = DOMPurify.sanitize(rendered);
+  return `<div class="chat-markdown">${cleanHtml}</div>`;
+}
+
+void import('./private-chat-dialog.component').then(
+  ({ PrivateChatDialogComponent }) => {
```
DenizAltunkapan: This dynamic import plus prototype patching is the part I'd push back on hardest. It swaps out formatMessage on PrivateChatDialogComponent asynchronously, as a side effect of chat-reactions.ts being imported somewhere. That means there's a real race: if the component gets instantiated and renders a message before this import resolves, it silently falls back to the old linkifyMessage behaviour with no markdown and no DOMPurify pass. Nothing in the test suite exercises that timing, and it would be very easy to miss in normal usage since the patch usually lands fast enough locally.

I'd rather see formatMessage changed directly inside private-chat-dialog.component.ts to call into renderChatMarkdown. That removes the load order dependency entirely and makes the data flow obvious to anyone reading the component.

frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
```diff
@@ -0,0 +1,45 @@
...
+    const prototype = PrivateChatDialogComponent.prototype as unknown as {
+      formatMessage: (content?: string) => string;
+    };
```
DenizAltunkapan: The original formatMessage returns SafeHtml, which is what lets the DomSanitizer trust the value handed to innerHTML in the template. This patched version returns a plain string but is assigned onto a property typed as SafeHtml, so Angular ends up trusting a value that never actually went through DomSanitizer.bypassSecurityTrustHtml. In practice that means the only thing standing between a message and the DOM is the DOMPurify.sanitize call in renderChatMarkdown, none of Angular's own sanitization layer is in play here anymore. DOMPurify is solid, but removing a defense in depth layer like this should be a deliberate, visible decision, not an incidental side effect of a type being widened during a monkey patch.

Once formatMessage is changed directly rather than patched, this becomes straightforward to fix: sanitize with DOMPurify as you do now, then wrap the result with this.sanitizer.bypassSecurityTrustHtml before returning it, same as the code this replaces effectively did.

frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
```diff
@@ -0,0 +1,45 @@
...
+    prototype.formatMessage = (content = '') => renderChatMarkdown(content);
```
DenizAltunkapan: Separate from the safety issue, this also throws away the formattedMessages cache that the original method relies on. There's a comment right above the old formatMessage explaining that the cache exists specifically so change detection doesn't re-parse and re-sanitize every visible message on every cycle. The replacement recomputes markdown and reruns DOMPurify on every call, so that performance work is quietly undone. Worth preserving once this moves into the component directly.

DenizAltunkapan (changes requested):
Thanks for tackling this, markdown support is a nice addition and the DOMPurify test coverage for the obvious XSS cases is good to see. I don't think this is ready to merge as is though, mainly because of how formatMessage gets replaced at runtime instead of being changed directly. Left inline notes on the specific spots, happy to take another look once those are addressed.

frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
prashantpiyush1111: Addressed. `formatMessage` is now implemented directly in `private-chat-dialog.component.ts` and calls `renderChatMarkdown` synchronously. The runtime prototype patch and `chat-reactions.ts` side-effect import have been removed, so there is no load-order race.

frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
prashantpiyush1111: Addressed. The direct `formatMessage` implementation sanitizes via `renderChatMarkdown` and then wraps the cleaned HTML with `this.sanitizer.bypassSecurityTrustHtml(...)`, preserving the original `SafeHtml` contract and Angular's sanitization boundary.

frontend/src/app/pages/private-chat-dialog/chat-markdown.ts
prashantpiyush1111: Addressed. The existing `formattedMessages` cache is preserved in `formatMessage`, keyed by message text, so repeated change-detection calls reuse the same `SafeHtml` result instead of re-parsing and re-sanitizing.

frontend/src/app/pages/private-chat-dialog/private-chat-dialog.component.ts
```diff
@@ -752,71 +752,26 @@ export class PrivateChatDialogComponent
...
       return '';
     }
+
     const cached = this.formattedMessages.get(content);
     if (cached !== undefined) {
       return cached;
     }
-    const formatted = this.linkifyMessage(content);
+
+    const cleanHtml = renderChatMarkdown(content);
```
DenizAltunkapan, line 769: The old linkifyMessage code rendered auto linked URLs with target="_blank" and rel="noopener noreferrer" on purpose, and private-chat-dialog.component.spec.ts still has tests asserting both of those attributes. Marked does not add either one to the links it generates, so a plain URL pasted into a chat message now opens in the same tab instead of a new one, which is a real step back for a chat window people expect to keep open while they follow a link.

This is not a security hole since DOMPurify still sanitises the output either way, but it is a behaviour regression against what the team explicitly wanted before, and the existing spec for it would fail if it ran. Worth noting separately, the frontend CI job only runs the build and the Prettier check, ng test never executes, so this did not get caught by the pipeline. That is a separate, pre-existing gap, not something to fix in this PR, but it does mean this one needs an actual look rather than relying on the green check.

Could you configure marked to add target="_blank" and rel="noopener noreferrer" to generated links (a custom renderer for link tokens, or a post process step before DOMPurify) and then run the existing spec locally to confirm it passes again.

DenizAltunkapan (changes requested):
Good progress here. The runtime patching is gone, formatMessage now calls renderChatMarkdown directly, the SafeHtml contract is respected again through bypassSecurityTrustHtml, and the formattedMessages cache is back in use with tests covering both the wiring and the cache reuse. That addresses everything from the previous round.

One thing still needs attention before this merges though, left as an inline note on where it happens.

DenizAltunkapan (approved):
@prashantpiyush1111 thanks!

PR #334, opened 2026-08-31
fix(auth): validate stored JWT server-side before allowing protected …

### LINKED ISSUE #330: fix(auth): validate stored JWT server-side before allowing protected navigation
## Problem

The frontend currently decides whether a stored authentication token is valid by decoding the JWT locally and checking only its `exp` claim. `authGuard` uses this result to allow navigation to protected pages.

This means a token can remain locally accepted by the frontend as long as it is structurally decodable and not past its `exp`, even if the backend has already invalidated or revoked that session/token.

## Expected behavior

Protected-route authentication should ultimately rely on server-side token/session validity rather than treating a locally decoded JWT as sufficient proof of an active session.

When the stored token/session is no longer valid, the frontend should clear the stale authentication state and redirect the user to login instead of continuing to consider the session authenticated.

## Current implementation

`AuthService.isLoggedIn()` reads the token from `localStorage`, decodes its payload, and returns `true` when the token has no `exp` claim or when `Date.now() < payload.exp * 1000`. `authGuard` then uses this boolean to permit navigation.

## Suggested direction

Reuse the application's existing refresh/authentication endpoint or another authenticated backend check to validate the session, while keeping route-guard behavior fail-closed when the server rejects the session.

This should also include tests covering a locally non-expired token that has become invalid server-side.

hazelr125:
## Summary
Fixed authentication guard to validate sessions server-side instead of relying solely on locally-decoded JWT expiry. Frontend now checks with backend before allowing navigation to protected routes, ensuring revoked/invalidated sessions are properly rejected.

### Changes
- `AuthService.validateTokenWithServer()`: Calls `/auth/refresh` to validate session server-side
- `AuthService.isLoggedInLocally()`: Extracted local JWT check (fast, client-side only)
- `authGuard`: Made async; validates with server before allowing navigation
- Added tests covering locally-valid, server-revoked token scenario

## Linked issue
Closes #330 

## How to test
1. Login with credentials
2. Manually revoke the session:
   ```sql
   UPDATE refresh_token SET revoked = true WHERE user_id = <your_user_id>;
3. Try to navigate to a protected route (e.g., /dashboard)
4. Expected: Frontend allows local check to pass, backend rejects refresh, user redirected to login
5. Verify: localStorage is cleared after redirect

## Notes / Risk
No breaking changes to existing endpoints
Backward compatible: isLoggedIn() still returns local check only for other use cases

### DIFF
--- a/frontend/src/app/auth.guard.spec.ts
+++ b/frontend/src/app/auth.guard.spec.ts
@@ -1,0 +1,119 @@
+import { TestBed } from '@angular/core/testing';
+import { Router } from '@angular/router';
+import { provideHttpClient } from '@angular/common/http';
+import {
+  HttpTestingController,
+  provideHttpClientTesting,
+} from '@angular/common/http/testing';
+import { authGuard } from './auth.guard';
+import { AuthService } from './services/auth.service';
+import { environment } from '../environments/environment';
+
+describe('authGuard', () => {
+  let httpMock: HttpTestingController;
+  let authService: AuthService;
+  let router: Router;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        provideHttpClient(),
+        provideHttpClientTesting(),
+        AuthService,
+        {
+          provide: Router,
+          useValue: { navigate: jasmine.createSpy('navigate') },
+        },
+      ],
+    });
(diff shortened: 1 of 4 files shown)


DenizAltunkapan (approved):
@hazelr125 thank you!

PR #335, opened 2026-09-05
Feature issue#155

### LINKED ISSUE #155: [Feature]: Chat Message Deletion
Users should be able to delete their own chat messages.
This feature introduces backend deletion logic and frontend integration with a clear user interaction.

**Backend Scope:**

* Add an endpoint to delete a chat message
* Ensure users can only delete their own messages
* Define deletion strategy (soft delete preferred)
* Persist deletion state and hide deleted content from standard queries
* Ensure chat history consistency

**Frontend Scope:**

* Add a delete action (trash icon or context menu) for own messages
* Prompt the user for confirmation before deletion
* Update the chat UI immediately after deletion
* Optionally show a placeholder (e.g. “Message deleted”)

AnUbH75:
## Summary

Implemented chat message deletion for private and group chats using the existing `clientMessageId`.

### Changes

* Added soft deletion using the `deleted` flag on `ChatMessage`.
* Updated message history queries to exclude deleted messages.
* Added sender authorization so users can only delete their own messages.
* Added the `/app/chat.delete` WebSocket handler.
* Added dedicated WebSocket destinations for deletion events:

  * `/user/queue/private/deleted`
  * `/topic/group/{groupId}/deleted`
* Added `ChatMessageDeletedDto` on both backend and frontend.
* Added frontend handling for deleted messages.
* Added a **hover-based delete UI** for the current user's messages:

  * When hovering over your own message, a delete option appears near the message timestamp.
  * Clicking the option opens a confirmation prompt before deletion.
* Added backend and frontend tests covering deletion, authorization, broadcasting, persistence, and frontend event handling.

## Linked issue

Closes #155

### Backend

```bash
cd backend
.\mvnw.cmd test
```

### Frontend

```bash
cd frontend
npm test
```

All tests are passing.

### Manual verification

1. Send a message in a private chat.
2. Hover over your own message and verify the delete option appears near the timestamp.
3. Delete the message and confirm the action.
4. Verify the message is removed/marked as deleted for both participants.
5. Repeat the same flow in a group chat.
6. Refresh/reopen the conversation and verify deleted messages do not reappear.
7. Verify the delete option is only available for messages sent by the current user.
8. Cancel the confirmation and verify the message remains unchanged.

## Notes / Risk

Deletion is implemented as a soft delete rather than physically removing the database record.

The existing schema update configuration handles the new `deleted` field.

Dedicated WebSocket deletion destinations keep deletion events separate from normal chat message subscriptions.

### DIFF
--- a/backend/src/main/java/vaultWeb/controllers/ChatController.java
+++ b/backend/src/main/java/vaultWeb/controllers/ChatController.java
@@ -11,6 +11,7 @@ import org.springframework.messaging.handler.annotation.Payload;
 import org.springframework.messaging.simp.SimpMessagingTemplate;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.stereotype.Controller;
+import vaultWeb.dtos.ChatMessageDeletedDto;
 import vaultWeb.dtos.ChatMessageDto;
 import vaultWeb.exceptions.UnauthorizedException;
 import vaultWeb.models.ChatMessage;
@@ -95,4 +96,39 @@ public class ChatController {
         responseDto.getSenderUsername(),
         responseDto.getPrivateChatId());
   }
+
+  @MessageMapping("/chat.delete")
+  public void deleteMessage(@Payload String clientMessageId, Principal principal) {
+
+    if (principal == null || principal.getName() == null || principal.getName().isBlank()) {
+      throw new UnauthorizedException("User not authenticated");
+    }
+
+    ChatMessage deletedMessage =
+            chatService.deleteMessage(
+                    clientMessageId,
+                    principal.getName());
+
+    ChatMessageDeletedDto responseDto =
+            new ChatMessageDeletedDto(deletedMessage.getClientMessageId());
+
+    if (deletedMessage.getGroup() != null) {
+      messagingTemplate.convertAndSend(
+              "/topic/group/" + deletedMessage.getGroup().getId(),
+              responseDto);
+    } else if (deletedMessage.getPrivateChat() != null) {
+      String user1 = deletedMessage.getPrivateChat().getUser1().getUsername();
+      String user2 = deletedMessage.getPrivateChat().getUser2().getUsername();
+
+      Set<String> recipients = new LinkedHashSet<>();
+      recipients.add(user1);
+      recipients.add(user2);
+
+      recipients.forEach(
+              user ->
+                      messagingTemplate.convertAndSendToUser(
+                              user, "/queue/private", responseDto));
+    }
+  }
+
 }
--- a/backend/src/main/java/vaultWeb/controllers/ChatController.java
+++ b/backend/src/main/java/vaultWeb/controllers/ChatController.java
@@ -47,7 +47,7 @@ public class ChatController {
     ChatMessageDto responseDto = chatService.toDto(savedMessage);
 
     messagingTemplate.convertAndSend(
-        "/topic/group/" + savedMessage.getGroup().getId(), responseDto);
+            "/topic/group/" + savedMessage.getGroup().getId(), responseDto);
   }
 
   private void authorizeGroupMessage(ChatMessageDto messageDto, Principal principal) {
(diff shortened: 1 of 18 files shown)


AnUbH75:
Please check the PR

backend/src/main/java/vaultWeb/controllers/ChatController.java
```diff
@@ -95,4 +96,33 @@ public void sendPrivateMessage(@Valid @Payload ChatMessageDto messageDto) {
...
   }
+
+  @MessageMapping("/chat.delete")
+  public void deleteMessage(@Payload String clientMessageId, Principal principal) {
+
+    if (principal == null || principal.getName() == null || principal.getName().isBlank()) {
+      throw new UnauthorizedException("User not authenticated");
+    }
+
+    ChatMessage deletedMessage = chatService.deleteMessage(clientMessageId, principal.getName());
```
DenizAltunkapan, line 111: This can throw EntityNotFoundException or AccessDeniedException, and line 104 above can throw UnauthorizedException. None of these are caught anywhere for STOMP handlers. GlobalExceptionHandler is a @ControllerAdvice for @RestController only, it does not apply to @MessageMapping methods. There is also no @MessageExceptionHandler or StompSubProtocolErrorHandler registered in WebSocketConfig. So if a delete fails (already deleted, not the sender, stale client state), the exception is only logged server side and the client gets no error frame, the message just silently stays visible. Please add a @MessageExceptionHandler (or equivalent) for this controller so delete failures are reported back to the client.

frontend/src/app/pages/private-chat-dialog/private-chat-dialog.component.html
```diff
@@ -136,21 +136,42 @@
...
+                [innerHTML]="formatMessage(msg.content)"
+              ></div>
+            </ng-container>
+          </ng-template>
+
+          <div class="message-time-row flex items-baseline gap-2 self-end mt-1">
+            <button
+              *ngIf="msg.senderUsername === currentUsername && !msg.deleted"
+              type="button"
+              class="delete-message-inline text-[10px] opacity-0 group-hover:opacity-70 hover:!opacity-100 transition-opacity"
```
DenizAltunkapan: The delete action is only revealed via group-hover, which means it is unreachable on touch devices with no hover state. Users on mobile/tablet would have no way to delete their own messages. Please add a touch-accessible way to trigger delete (e.g. always visible on small viewports, or a long-press/tap affordance).

DenizAltunkapan (changes requested):
Solid implementation overall (soft delete, sender authorization, dedicated deletion topics with a regression test). Left two blocking review points inline before this is ready to merge.

AnUbH75:
Fixed both of these in the latest push:

**STOMP failures now reported back to the client** — Added a `@MessageExceptionHandler` in `ChatController` catching `EntityNotFoundException`, `AccessDeniedException`, `UnauthorizedException`, and `IllegalArgumentException` from this controller's `@MessageMapping` methods, relaying them via `@SendToUser("/queue/errors")` as a `ChatErrorDto`. Frontend subscribes to `/user/queue/errors` and surfaces it as a toast, so a failed delete (already deleted, not the sender, etc.) is no longer silent.

One caveat: for `UnauthorizedException` specifically, `principal` is `null` by definition (that's the condition that throws it), so `@SendToUser` has no session to route to in that case — it'll still only be server-logged. That's a limitation of STOMP's user-destination routing rather than something this handler can work around. The other three cases (which cover the actual delete-failure scenarios you flagged) are now correctly reported back.

**Delete option now reachable on touch devices** — Added `pointer-coarse:opacity-70` to the button's classes. That's Tailwind's variant for `@media (pointer: coarse)`, matching touch screens but not mouse/trackpad, so:
- Touch devices: delete option always visible, no hover required
- Mouse/trackpad: unchanged, still hover-to-reveal

Went with capability-detection over a viewport-width breakpoint since it felt more correct (e.g. a touch laptop with a large screen still needs this).

@DenizAltunkapan 

---

DenizAltunkapan (approved):

PR #345, opened 2026-09-09
Feature issue#341

### LINKED ISSUE #341: [Bug]: Private chat WebSocket handler allows sender impersonation
## What happened?

The `sendPrivateMessage` WebSocket handler accepts `senderId` and `senderUsername` directly from the client payload instead of deriving them from the authenticated STOMP principal. This allows any connected client to impersonate another user and inject messages into any private chat by ID.

The group chat equivalent already guards against this: `authorizeGroupMessage` checks membership and overwrites `senderUsername` from `principal.getName()` before the message is saved. The private chat handler has no such check.

## Steps to reproduce

1. Open a WebSocket/STOMP connection as an authenticated user.
2. Send a payload to `/app/chat.private.send` with an arbitrary `privateChatId` (one you are not a participant of) and a `senderUsername` set to another user's username.
3. Observe that `ChatService.saveMessage` persists and broadcasts the message as if it came from the spoofed sender, to both real participants of that private chat.

## Expected

`senderUsername` (and `senderId`) should be derived from the authenticated principal on the STOMP session, the same way `sendMessage` does for group chats, and the caller's membership in the target private chat should be verified before saving.

## Actual

`sendPrivateMessage` has no `Principal` parameter at all and passes the client-supplied `ChatMessageDto` straight to `chatService.saveMessage`, trusting `senderId`/`senderUsername` as sent.

## Evidence

- Logs/error message: Logic in `ChatController.java`:

```java
@MessageMapping("/chat.private.send")
public void sendPrivateMessage(@Valid @Payload ChatMessageDto messageDto) {
  ChatMessage savedMessage = chatService.saveMessage(messageDto);
  ...
}
```

Compare to the group chat path in the same file, which resolves the sender from the security context:

```java
String username = principal.getName();
boolean isMember =
    groupMemberRepository.existsByGroupIdAndUserUsername(messageDto.getGroupId(), username);
if (!isMember) {
  throw new AccessDeniedException("Not allowed to send messages to this group");
}
messageDto.setSenderId(null);
messageDto.setSenderUsername(username);
```

This is the same class of bug that was fixed for the REST endpoint in #314 (sender impersonation in `PrivateChatController`), but the fix was not applied to this WebSocket path.

## Environment

- Where: dev
- Browser/OS: N/A (WebSocket/STOMP API)
- Feature flag(s) (if any): none

AnUbH75:
## Summary

`sendPrivateMessage` (WebSocket handler for `/app/chat.private.send`) trusted `senderId`/`senderUsername` directly from the client payload and had no check that the caller was actually a participant of the target `privateChatId`. This allowed any authenticated client to impersonate another user and inject messages into any private chat by ID — the same class of bug fixed for the REST layer in #314, but never applied to this WebSocket path.

Fixed by mirroring the existing group-chat authorization pattern (`authorizeGroupMessage`) with a new `authorizePrivateMessage`:
- Added `Principal principal` param to `sendPrivateMessage`
- Added `PrivateChatRepository.existsByIdAndParticipantUsername(id, username)` to check the caller is `user1` or `user2` on the target private chat
- `senderId`/`senderUsername` are now always overwritten from `principal.getName()` before the message reaches `ChatService.saveMessage`, discarding whatever the client sent
- Unauthorized sends now throw `AccessDeniedException` before anything is persisted or broadcast

## Linked issue

Closes #341

## How to test

**Unit tests** (added to `ChatControllerTest`):
- `shouldSendPrivateMessage_WhenAuthenticatedUserIsParticipant` — happy path, both participants receive the broadcast
- `shouldRejectPrivateMessage_WhenAuthenticatedUserIsNotParticipant` — regression test for the exploit: a non-participant is rejected, nothing is saved or sent
- `shouldRejectPrivateMessage_WhenUnauthenticated` — no principal → `UnauthorizedException`
- `shouldOverwriteSenderUsername_WithPrincipalName_NotSpoofedPayloadValue` — confirms a spoofed `senderUsername`/`senderId` in the payload never reaches `saveMessage`

Run: `./mvnw test -Dtest=ChatControllerTest` (or full suite, since `PrivateChatRepository` and `ChatController`'s constructor both changed).

**Manual/STOMP verification** (repro from the issue, now blocked):
1. Connect as an authenticated user who is *not* a participant of a target private chat
2. `SEND` to `/app/chat.private.send` with that chat's `privateChatId` and `senderUsername` set to a real participant
3. Expect rejection at the handler and confirm no row is written to `chat_message`

Also manually smoke-tested normal private chat send/receive between two real participants, and group chat send, to confirm no regression in either path.

## Notes / Risk

- No schema migration — only a new repository query method (JPQL `@Query`, no new column/index).
- No feature flag; this is a straight security fix with no behavior change for legitimate senders.
- `ChatController`'s constructor gained a new dependency (`PrivateChatRepository`) — anything else that mocks/constructs `ChatController` directly (rather than via Spring) will need to supply it.
- `existsByIdAndParticipantUsername` returns `false` uniformly for both "chat doesn't exist" and "caller isn't a participant," so the client can't distinguish the two from the error — this is intentional (avoids leaking chat existence) but worth being aware of if any future caller wants a different error for a bad ID vs. a permission failure.

### DIFF
--- a/backend/src/main/java/vaultWeb/controllers/ChatController.java
+++ b/backend/src/main/java/vaultWeb/controllers/ChatController.java
@@ -23,6 +23,12 @@ import vaultWeb.repositories.GroupMemberRepository;
 import vaultWeb.repositories.PrivateChatRepository;
 import vaultWeb.services.ChatService;
 
+/**
+ * Controller responsible for handling WebSocket-based chat functionality.
+ *
+ * <p>Supports both group chat and private messages. Messages are first persisted via ChatService
+ * and then dispatched to the corresponding topics or users.
+ */
 @Slf4j
 @Controller
 @RequiredArgsConstructor
@@ -33,6 +39,12 @@ public class ChatController {
   private final GroupMemberRepository groupMemberRepository;
   private final PrivateChatRepository privateChatRepository;
 
+  /**
+   * Handles incoming group chat messages from clients and broadcasts them to all subscribers of the
+   * specified group topic.
+   *
+   * @param messageDto DTO containing message content, sender information, and target group
+   */
   @MessageMapping("/chat.send")
   public void sendMessage(@Valid @Payload ChatMessageDto messageDto, Principal principal) {
     authorizeGroupMessage(messageDto, principal);
@@ -63,6 +75,12 @@ public class ChatController {
     messageDto.setSenderUsername(username);
   }
 
+  /**
+   * Handles incoming private chat messages from clients and sends them to both users of the private
+   * chat. The message content is end-to-end encrypted and never decrypted by the server.
+   *
+   * @param messageDto DTO containing message content, sender information, and private chat ID
+   */
   @MessageMapping("/chat.private.send")
   public void sendPrivateMessage(@Valid @Payload ChatMessageDto messageDto, Principal principal) {
     authorizePrivateMessage(messageDto, principal);
--- a/backend/src/main/java/vaultWeb/controllers/ChatController.java
+++ b/backend/src/main/java/vaultWeb/controllers/ChatController.java
@@ -15,14 +15,9 @@ import vaultWeb.dtos.ChatMessageDto;
 import vaultWeb.exceptions.UnauthorizedException;
 import vaultWeb.models.ChatMessage;
 import vaultWeb.repositories.GroupMemberRepository;
+import vaultWeb.repositories.PrivateChatRepository;
 import vaultWeb.services.ChatService;
 
-/**
- * Controller responsible for handling WebSocket-based chat functionality.
- *
- * <p>Supports both group chat and private messages. Messages are first persisted via ChatService
- * and then dispatched to the corresponding topics or users.
- */
 @Slf4j
 @Controller
 @RequiredArgsConstructor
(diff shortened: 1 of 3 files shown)


AnUbH75:
@DenizAltunkapan pls have a look

backend/src/main/java/vaultWeb/controllers/ChatController.java
```diff
@@ -20,14 +20,9 @@
...
 import vaultWeb.repositories.GroupMemberRepository;
+import vaultWeb.repositories.PrivateChatRepository;
 import vaultWeb.services.ChatService;
 
-/**
- * Controller responsible for handling WebSocket-based chat functionality.
- *
- * <p>Supports both group chat and private messages. Messages are first persisted via ChatService
- * and then dispatched to the corresponding topics or users.
- */
```
DenizAltunkapan, line 30: is there any reason why you removed the javadoc? @AnUbH75

AnUbH75: No intentional reason, that's a merge artifact from resolving the conflict with #155 — restoring it now

AnUbH75:
@DenizAltunkapan retrieved

DenizAltunkapan (approved):
@AnUbH75 thank you very much

PR #346, opened 2026-09-11
Feature issue#342

### LINKED ISSUE #342: [Bug]: GET /api/dashboard/{username} exposes any user's dashboard without authorization
## What happened?

The `GET /api/dashboard/{username}` endpoint returns aggregated dashboard data for any username without checking that the caller owns that username or has admin rights. Any authenticated user can pull another user's dashboard data by guessing or enumerating usernames.

## Steps to reproduce

1. Authenticate as any regular user and obtain a valid JWT.
2. Send `GET /api/dashboard/{otherUsername}` for a username that is not your own.
3. Observe that the response contains the target user's dashboard data instead of a 403.

## Expected

The endpoint should either be removed in favor of `/api/dashboard/me`, or restricted to admin callers only, with a check that rejects requests where the caller is neither the target user nor an admin.

## Actual

`getDashboardForUser` looks up the target user directly from the path variable and returns their dashboard with no ownership or role check, unlike other admin-facing endpoints in the codebase which are explicitly gated.

## Evidence

- Logs/error message: Logic in `DashboardController.java`:

```java
@GetMapping("/{username}")
@Operation(
    summary = "Get dashboard data for a specific user",
    description = "Primarily useful for admin tooling or debugging in the UI.")
public ResponseEntity<UserDashboardDto> getDashboardForUser(@PathVariable String username) {
  User targetUser =
      userRepository
          .findByUsername(username)
          .orElseThrow(() -> new UserNotFoundException("User not found: " + username));
  return ResponseEntity.ok(dashboardService.buildDashboard(targetUser));
}
```

The docstring says this is "primarily useful for admin tooling," but there is no admin check enforcing that. It exposes the target user's group memberships and roles, private chat partner usernames, poll participation, and message counts to any authenticated caller.

## Environment

- Where: dev
- Browser/OS: N/A (REST API)
- Feature flag(s) (if any): none

AnUbH75:
## Summary

`GET /api/dashboard/{username}` had no authorization check — `getDashboardForUser` looked up any requested username and returned their full dashboard (group memberships/roles, private chat partner usernames, poll participation, message counts) with no verification that the caller owned that username or had elevated privileges.

Investigated whether an existing admin mechanism could gate this instead of removing the endpoint's cross-user capability entirely. Found that this codebase's only role concept (`Role.ADMIN`/`Role.USER`) is scoped per-`GroupMember` — enforced via `@AdminOnly`/`AdminOnlyAspect`, which requires a `groupId` to check a user's role *within that specific group*. There is no global-admin concept anywhere in the system that could legitimately gate "can user X view user Y's unrelated dashboard." Inventing one wasn't appropriate scope for a security fix.

Fixed by restricting `getDashboardForUser` to self-only access: it now resolves the caller's identity via `authService.getCurrentUser()` (the same trusted source `/api/dashboard/me` already uses) and rejects the request unless the caller's own username matches the requested path variable.

This makes the endpoint functionally redundant with `/api/dashboard/me` — flagging that explicitly below rather than silently leaving overlapping endpoints in place.

## Linked issue

Closes #342

## How to test

**Unit tests** (`DashboardControllerTest`):
- `shouldReturnDashboardForCurrentUser` — existing, unaffected
- `shouldReturnDashboardForSpecificUser_WhenRequestingOwnUsername` — updated from the original (which never stubbed `getCurrentUser()`, since the old code never called it) to reflect the new self-check
- `shouldRejectDashboard_WhenRequestingAnotherUsersUsername` — **regression test**: an authenticated user requesting a different username gets `AccessDeniedException`, and neither the repository lookup nor `buildDashboard` is ever invoked
- `shouldRejectDashboard_WhenUnauthenticated` — no current user → `UnauthorizedException`
- `shouldThrowUserNotFound_WhenOwnUsernameLookupFails` — self-check passes but repository lookup misses (stale session edge case)

Run: `./mvnw test -Dtest=DashboardControllerTest`

**Manual verification** (repro from the issue, now blocked):
1. Authenticate as user A, obtain a JWT
2. `GET /api/dashboard/{userB}` for a username that isn't A's own
3. Expect `403 Forbidden` instead of user B's dashboard data
4. Confirm `GET /api/dashboard/{ownUsername}` and `GET /api/dashboard/me` both still return correctly for the authenticated user

## Notes / Risk

- No schema migration — no new fields, no new tables.
- No feature flag; straight security fix, no behavior change for legitimate self-access.
- **Breaking change for any existing client** relying on `GET /api/dashboard/{username}` to view *another* user's dashboard for legitimate admin/debugging purposes — that capability is fully removed, not just gated, since no real authorization boundary exists to gate it on. If admin tooling genuinely needs cross-user dashboard access, that requires a separate design discussion (introducing a real global-admin role) rather than this fix.
- Endpoint is now redundant with `/api/dashboard/me` (same result, reached via path variable instead of implicit caller identity) — worth a maintainer call on whether to deprecate/remove `getDashboardForUser` entirely in a follow-up rather than keep both.
- Confirm `AccessDeniedException` is already mapped to a 403 response globally (Spring Security default, or via `GlobalExceptionHandler`) — not verified against the actual handler config in this PR.

### DIFF
--- a/backend/src/main/java/vaultWeb/controllers/DashboardController.java
+++ b/backend/src/main/java/vaultWeb/controllers/DashboardController.java
@@ -4,6 +4,7 @@ import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.AccessDeniedException;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -40,8 +41,21 @@ public class DashboardController {
   @GetMapping("/{username}")
   @Operation(
       summary = "Get dashboard data for a specific user",
-      description = "Primarily useful for admin tooling or debugging in the UI.")
+      description =
+          "Returns dashboard data only when the requested username matches the authenticated "
+              + "caller. There is no global-admin role in this system to gate broader access "
+              + "on (roles here are scoped per-group via GroupMember/@AdminOnly), so this "
+              + "endpoint cannot safely support cross-user lookups. Prefer /api/dashboard/me.")
   public ResponseEntity<UserDashboardDto> getDashboardForUser(@PathVariable String username) {
+    User currentUser = authService.getCurrentUser();
+    if (currentUser == null) {
+      throw new UnauthorizedException("User is not authenticated");
+    }
+
+    if (!currentUser.getUsername().equals(username)) {
+      throw new AccessDeniedException("Not allowed to view this user's dashboard");
+    }
+
     User targetUser =
         userRepository
             .findByUsername(username)
--- a/backend/src/test/java/vaultWeb/controllers/DashboardControllerTest.java
+++ b/backend/src/test/java/vaultWeb/controllers/DashboardControllerTest.java
@@ -1,6 +1,10 @@
 package vaultWeb.controllers;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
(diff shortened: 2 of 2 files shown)


DenizAltunkapan (approved):
@AnUbH75 thanks

PR #355, opened 2026-09-15
Feature issue#281

### LINKED ISSUE #281: Add authenticated same-origin proxy for workspace file previews
## Problem

The Cloud Page inline viewer endpoint requires bearer-token authentication. Native `<video>`, `<audio>`, ``, and embedded PDF elements cannot attach an arbitrary `Authorization` header, so the frontend cannot use the endpoint directly while preserving native media streaming and byte-range seeking.

## Proposed solution

Add an authenticated same-origin frontend route or proxy that:

- authenticates the current Vault Web user using the existing frontend/session flow
- forwards requests to `GET /api/files/view` with the backend bearer token
- preserves `Range`, conditional request, and relevant response headers
- streams the response without buffering the whole file
- never accepts or exposes paths outside the authenticated user's root

## Acceptance criteria

- [ ] Images, audio, video, and PDFs can use a same-origin URL from native browser elements.
- [ ] Byte-range requests and `206 Partial Content` responses are preserved.
- [ ] `Content-Type`, `Content-Disposition`, `ETag`, `Last-Modified`, `Content-Range`, and `Accept-Ranges` are forwarded where applicable.
- [ ] Authentication tokens are not placed in query strings or exposed to the browser media URL.
- [ ] Unauthorized and out-of-root requests fail without exposing file contents.

Related to Vault-Web/cloud-page#98 and Vault-Web/cloud-page#102.

AnUbH75:
## Summary
Native <video>/<audio>/ playback previously went through a full blob download
(`getFileView()` → fetch entire file → `URL.createObjectURL`), so playback couldn't start
until the whole file arrived and seeking wasn't possible. Video/audio had no preview path at
all and just forced a download.

This adds a same-origin streaming proxy so the browser's native media elements can play and
seek files directly, backed by real HTTP Range support.

- `MediaTokenService` — mints short-lived (default 5 min), single-file-scoped tokens, signed
  with a **separate key** from `jwt.secret`. This is deliberate: jwt.secret is shared with
  cloud-page, and a token traveling in a URL must never double as a general access token for
  either service.
- `CloudPageMediaClient` — forwards a request to cloud-page's `/api/files/view` on behalf of a
  user, passing through `Range`/`If-None-Match`/`If-Modified-Since`, streaming the response
  body rather than buffering it.
- `MediaController` — exposes `POST /api/media/token` (authenticated, mints a token for one
  file) and `GET /api/media/stream/{token}` (no auth header required/possible — the token in
  the path is the credential, scoped to exactly one file).
- Frontend: `previewFile()` now covers video/audio and routes all previewable types through
  the streaming URL instead of a blob fetch. Opens the destination tab synchronously to avoid
  popup-blocker issues with the async token round-trip.

## Linked issue

Closes #281

Depends on the cloud-page side of this (Range support added to `/api/files/view`) — see
Vault-Web/cloud-page#<fill in the PR number from the cloud-page PR>. Without that change this
proxy has nothing to actually forward a `206` from.

## How to test
.\mvnw test "-Dtest=MediaTokenServiceTest,MediaControllerTest"

Manual end-to-end:
1. Log in, upload a video, click it — should open in a new tab and play immediately rather
   than waiting for a full download
2. Drag the video's progress bar — should seek instantly; DevTools Network tab should show
   `206` responses with `Content-Range` headers
3. Wait past the token TTL (or temporarily lower `app.media.token-ttl-seconds`), retry the
   same stream URL — should get `401`

## Notes / Risk

- Requires two new properties: `jwt.mediaSecret` (a **different** random value from
  `jwt.secret` — do not reuse it) and `app.cloudpage.base-url`. `app.cloudpage.trust-self-signed`
  is for local dev only, where cloud-page runs on a self-signed cert — should be `false` in
  any real deployment.
- `SecurityConfig` needed one new `permitAll` entry: `/api/media/stream/**` only. The token
  mint endpoint (`/api/media/token`) stays behind normal authentication.
- Rate limiting: cloud-page's `/api/files/view` shares its rate-limit bucket with regular
  downloads (`RateLimitCategory.DOWNLOAD`, 120/min per user). Aggressive scrubbing generates a
  burst of Range requests against that same budget — not addressed in this PR, flagged as a
  possible follow-up if it proves to be a real problem in practice.
- Verbose Tomcat-level `IOException: connection aborted` logging shows up on every seek/tab
  close, since the browser doesn't cleanly finish the in-flight chunk. This is expected
  streaming-proxy behavior (the code's own try/catch already handles it; the extra log line
  comes from Tomcat's own async-request logging, not from unhandled application code) and
  hasn't been suppressed here.


AnUbH75:
@DenizAltunkapan pls have a look on this

DenizAltunkapan (approved):

